PR #3, opened 2023-01-27
Add Atlas theme

ajlende:
Kicking things off with a variation I submitted for TT3 that wasn't selected: https://github.com/WordPress/twentytwentythree/pull/92

Screen Shot 2023-01-27 at 16 19 02-fullpage

---

Colors

---

Atlas Demo

---

I was messing around with some of the upcoming [CSS Color Module Level 5](https://www.w3.org/TR/css-color-5/) features for generating interesting gradients from single colors.

```CSS
background: linear-gradient(
    141deg,
    oklch(
        from var(--wp--preset--color--primary)
        calc( l * 0.75 )
        calc( c * 0.8 )
        calc( h + 12 )
    )
    var(--wp--preset--color--primary) 71%,
    oklch(
        from var(--wp--preset--color--primary)
        calc( l * 0.875 + 0.125 )
        calc( c * 0.8 )
        calc( h - 12 )
  ) 
);
```

Since there isn't much browser support, the calculated parts have been pre-computed with [Color.js](https://colorjs.io) which was created by the authors of the new spec, so it should be consistent with how the computed values would appear.

```ts
import atlas from "./atlas.json" assert { type: "json" };
import * as Color from "https://colorjs.io/src/index-fn.js";

Color.ColorSpace.register(Color.sRGB);
Color.ColorSpace.register(Color.OKLCH);

type Lightness = number;
type Chroma = number;
type Hue = number;
interface CSSLCHCoords {
  l: Lightness;
  c: Chroma;
  h: Hue;
}
interface LCHFn<T> {
  (v: CSSLCHCoords): T;
}
type LightnessFn = LCHFn<Lightness>;
type ChromaFn = LCHFn<Chroma>;
type HueFn = LCHFn<Hue>;

function namedLCHCoords([l, c, h]: [Lightness, Chroma, Hue]): CSSLCHCoords {
  return ({ l, c, h });
}

function oklchFrom(color: string, l: LightnessFn, c: ChromaFn, h: HueFn) {
  const s0 = color;
  const s1 = Color.parse(s0);
  const s2 = Color.to(s1, Color.OKLCH);
  const v = namedLCHCoords(s2.coords);
  s2.coords = [l(v), c(v), h(v)];
  const s3 = Color.toGamut(s2, {
    method: "oklch.c",
    space: Color.sRGB,
  });
  const s4 = Color.to(s3, Color.sRGB);
  const s5 = Color.serialize(s4, {
    format: "hex",
  });
  return s5;
}

const gradients = atlas.settings.color.palette.map(({ color, name, slug }) => {
  const dark = oklchFrom(
    color,
    ({ l }) => l * 0.75,
    ({ c }) => c * 0.8,
    ({ h }) => h + 12,
  );
  const light = oklchFrom(
    color,
    ({ l }) => l * 0.875 + 0.125,
    ({ c }) => c * 0.8,
    ({ h }) => h - 12,
  );

  return {
    gradient: `linear-gradient(141deg, ${dark}, ${color} 71%, ${light})`,
    name,
    slug,
  };
});

Deno.writeTextFileSync(
  "./atlas.json",
  JSON.stringify(
    {
      ...atlas,
      settings: {
        ...atlas.settings,
        color: {
          ...atlas.settings.color,
          gradients,
        },
      },
    },
    null,
    "\t",
  ),
);
```


atlas/style.css
```diff
@@ -0,0 +1,16 @@
+/*
+Theme Name: Atlas
+Theme URI: https://github.com/MaggieCabrera/community-themes/tree/trunk/atlas
+Author: ajlende
```
MaggieCabrera: I think both in here and the changelog this should be the WordPress team once we move this repo the WP namespace, but we can change that in a followup

MaggieCabrera:
Thanks for working on this @ajlende This is looking pretty good! Since this is now a child theme, what do you think about adding some templates that showcase some of the changes that the variation originally intended for the theme?

atlas/style.css
ajlende: Fixed in 0871b68 to match what I saw in core themes.

ajlende:
@MaggieCabrera I made a few more tweaks to the theme.json so it's in a place that I like. The only thing different from the screenshot is a thicker border on `:active` buttons, and links are only yellow when `:active` as well.

> what do you think about adding some templates that showcase some of the changes that the variation originally intended for the theme?

We talked about this over a video call. To summarize:

- Keeping with the Atlas—holding up the heavens—theme, it could be cool to include the [Starscape block](https://wordpress.org/plugins/starscape/) for a really flashy home page, and falling back to some static CSS or something if the plugin isn't present.
- Not all of the templates need to be modified, but there's something that I'd like to change on most of them.
- Anything that includes text should be added as a pattern so that it can be translated.
- It's fine to merge the existing theme.json without those template changes right now so they can be updated over time so it's more than just another TT3 variation.

MaggieCabrera (approved):
We can bring this in and keep working on other PRs. If the template changes are too many, we can decide if we want to branch out from TT3. Another reason to keep this theme as a child is that if we don't, we would need to bring all the other theme.json settings that, right now, this is inheriting from the parent. We can discuss it as we work on the theme. Great work so far! I'm excited about the stars block.

PR #13, opened 2023-03-09
Add new Stacks theme

scruffian:
This adds a new theme which can be used to create presentation decks. This idea is that the stacks are supplied in a pattern, which a user can add to a post or page and then use each post/page as a slide deck. The home page creates a demo which explains how it works.

The name "Stack" is already taken but "Stacks" seems to be free.

Design by @SaxonF


stacks/readme.txt
```diff
@@ -0,0 +1,40 @@
+=== Stacks ===
+Contributors: @SaxonF @scruffian
```
MaggieCabrera: According to [the guidelines](https://make.wordpress.org/themes/handbook/review/required/#9-files), this should be just one wp.org username, so maybe `wordpressdotorg` is best here. I'm not sure where is the best place to name the contributors. The [guidelines](https://make.wordpress.org/themes/handbook/review/required/#9-files) say 

> Themes can include one single front-facing credit link, which is restricted to the Theme URI or Author URI defined in style.css.

I think we should discuss what's the best way to do this and add that info to the readme file on this repo so we are consistent on all our themes

stacks/style.css
```diff
@@ -0,0 +1,16 @@
+/*
+Theme Name: Stacks
+Theme URI: 
+Author: Automattic
```
MaggieCabrera: Maybe this is where we add the contributor's names? but it might be very long for some themes and this is what it shows here:

Screenshot 2023-03-09 at 16 37 00

stacks/style.css
```diff
@@ -0,0 +1,16 @@
...
+Author URI: 
+Description: Stack your story
+Requires at least: 5.8
```
MaggieCabrera: This is different from what readme.txt says. Out of curiosity, is this coming from CBT?

stacks/style.css
```diff
@@ -0,0 +1,16 @@
...
+Tested up to: 5.9
+Requires PHP: 5.7
+Version: 0.0.1
+License: GNU General Public License v2 or later
+License URI: http://www.gnu.org/licenses/gpl-2.0.html
+Template:
```
MaggieCabrera: if it's not a child theme I think this can go away. Again, if this is from CBT, we should maybe file an issue so that it's not included when we are not creating a child theme

stacks/patterns/index.php
```diff
@@ -0,0 +1,109 @@
...
+ * Slug: stacks/index
+ * Categories: featured
+ */
+?>
+<!-- wp:group {"style":{"dimensions":{"minHeight":"100vh"},"position":{"type":"sticky","top":"0px"}},"backgroundColor":"white","layout":{"type":"flex","orientation":"vertical","verticalAlignment":"center","justifyContent":"center"}} -->
+<div class="wp-block-group has-white-background-color has-background" style="min-height:100vh"><!-- wp:group {"layout":{"type":"constrained"}} -->
+<div class="wp-block-group">
+
+<!-- wp:heading {"level":1,"fontSize":"x-large"} -->
+<h1 class="wp-block-heading has-x-large-font-size">Create Stacks.</h1>
```
MaggieCabrera: We should make all the strings translatable. I think @madhusudhand had a script for that somewhere? Could that be in a package we can use in this repo?

stacks/patterns/index.php
```diff
@@ -0,0 +1,109 @@
...
+<!-- /wp:heading --></div>
+<!-- /wp:group --></div>
+<!-- /wp:group -->
+
+<!-- wp:group {"style":{"dimensions":{"minHeight":"100vh"},"position":{"type":"sticky","top":"0px"},"color":{"background":"#f6f6f6"}},"layout":{"type":"flex","orientation":"vertical","verticalAlignment":"center","justifyContent":"center"}} -->
```
MaggieCabrera: Since the background colors of the slides are not part of the theme's palette, should we use something other than gray? this is a naive question from a non designer perspective, I'd just think it would look cooler. In the same way you added the image to the last slide, that made the whole theme much more solid, in my opinion. Would love other designer's eyes on this for feedback.

stacks/readme.txt
MaggieCabrera: I opened [this PR](https://github.com/WordPress/community-themes/pull/14) for discussion

stacks/style.css
scruffian: Yeah I think so

MaggieCabrera: Let's make it 6.1 or 6.2, since we need the sticky positioning

stacks/patterns/index.php
```diff
@@ -0,0 +1,109 @@
+<?php
+/**
+ * Title: Stacks
+ * Slug: stacks/index
```
jffng: Should this be called `stacks/home` and be named `home.php` instead, since it's used in the home template?

stacks/patterns/index.php
jffng: That function lives [here](https://github.com/Automattic/themes/blob/trunk/theme-utils.mjs#L1213) and is used as an [npm script](https://github.com/Automattic/themes/blob/trunk/package.json#L31)

Maybe we could start a paired down version of `theme-utils` in this repo, and eventually extract the shared utilities into an npm package?

stacks/patterns/index.php
```diff
@@ -0,0 +1,109 @@
...
+<div class="wp-block-cover"><span aria-hidden="true" class="wp-block-cover__background has-background-dim"></span><img class="wp-block-cover__image-background wp-image-29" alt="" src="<?php echo esc_url( get_stylesheet_directory_uri() ); ?>/assets/images/stack.jpg" style="object-position:0% 50%" data-object-fit="cover" data-object-position="0% 50%"/><div class="wp-block-cover__inner-container"><!-- wp:group {"layout":{"type":"flex","orientation":"vertical","verticalAlignment":"center","justifyContent":"center"}} -->
+<div class="wp-block-group"><!-- wp:heading {"textAlign":"center","textColor":"white","fontSize":"x-large"} -->
+<h2 class="wp-block-heading has-text-align-center has-white-color has-text-color has-x-large-font-size">Create your own Stacks</h2>
+<!-- /wp:heading -->
+
+<!-- wp:list {"ordered":true} -->
+<ol><!-- wp:list-item -->
+<li>Add the "stacks" pattern to a new post.</li>
```
jffng: I noticed this text is black in the site editor, but white on the front end. Any idea why? 

Screenshot 2023-03-09 at 10 32 04 PM

Also it shows up with padding but maybe that's a bigger site wide padding issue. 

Running Gutenberg trunk.

MaggieCabrera:
I wonder if a nav block with anchor links to each section would work here. Maybe adding a small script to smooth the scroll between jumps? When I have a moment I'm going to test that

carolinan:
"Stacks" doesnt relate to presentations or presentation decks to me. Is there any other name that would make it more intuitive to search for the theme's purpose?

scruffian:
> Is there any other name that would make it more intuitive to search for the theme's purpose?

Deck and pitch are unavailable, but slide(s) might be available.

scruffian:
> I wonder if a nav block with anchor links to each section would work here.

Where would it sit? We don't have many controls for z-index so I think it would get hidden behind the panels...

stacks/patterns/index.php
scruffian: Done

stacks/patterns/index.php
scruffian: I made a script for this here: https://github.com/WordPress/community-themes/pull/20

MaggieCabrera:
> > I wonder if a nav block with anchor links to each section would work here.
> 
> Where would it sit? We don't have many controls for z-index so I think it would get hidden behind the panels...

yeah, I gave it a try and it didn't work, we need a fixed position now!!!

stacks/style.css
scruffian: Done

stacks/style.css
scruffian: Done

stacks/patterns/index.php
scruffian: Updated the strings

stacks/patterns/index.php
scruffian: This is a Gutenberg issue I am investigating

scruffian:
Still outstanding on this are:
1. Should the slides use different colors?
2. What name should we go with?

MaggieCabrera:
> Still outstanding on this are:
> 
>     1. Should the slides use different colors?
> 
>     2. What name should we go with?

The colors could be a follow up if we get a designer's opinion on the matter. Shouldn't block this PR from landing.

The name, I'm not sure. I kinda like stacks, but it's again a subjective opinion.

MaggieCabrera (approved):

scruffian:
Ok I'm going to merge this, but I think it would still be good to get more opinions on the name and colors. Happy to change them after the merge.

PR #14, opened 2023-03-09
Global Readme: Theme guidelines

MaggieCabrera:
This PR expands the readme to include some guidelines for contributors and to serve as a reference for everyone that reviews themes in this repo.

This is not meant to be exhaustive (the wp.org theme repo guidelines are already there for a reason) but after some time working on themes we see some things pop up again and again and it's good to keep a reminder.

This PR is open for discussion as to what to include here. One thing that I would like other people's opinion is how to attribute credit to contributors for each specific theme. The Contributors line in the readme.txt of a theme should only have one wp.org username according to [the guidelines ](https://make.wordpress.org/themes/handbook/review/required/#9-files). What is the best place to add the names of everyone involved in each specific theme? Should we keep a contributors file or something like that?


scruffian:
> One thing that I would like other people's opinion is how to attribute credit to contributors for each specific theme. 

Can we add more meta to the readme.txt, for example:

Builders: @username

scruffian (approved):

Issue #6, opened 2023-02-15
[Blue Note]: Single template with sidebar


MaggieCabrera:
**Desktop:**

![Single (with sidebar)](https://user-images.githubusercontent.com/3593343/219082896-778f8b4d-924a-48aa-ae43-5b7782939cf9.jpg)

**Mobile:**

![Single (with sidebar) - Mobile](https://user-images.githubusercontent.com/3593343/219083059-55b9982d-6b10-41f5-934c-3ca124e2cd84.jpg)

mkwpdev:
[blue-note-child-single-template.zip](https://github.com/WordPress/community-themes/files/11688752/blue-note-child-single-template.zip)

carolinan:
How do we hide the author details + tags, and navigation on mobile?
I don't think the Gutenberg team is ready to decide whether there should be a way to hide blocks in the mobile view.
So we need to use custom CSS.

MaggieCabrera:
Let's forget about hiding the author/tags on mobile, we can keep them.

To do:

- [x] Post date should be in italics
- [x] Post title and date should be centered vertically
- [ ] Can placeholder images have aspect ratio applied to them?
- [x] The featured image needs a duotone filter. Wish we could control the filter from the post, else they all need to be the same color!
- [x] Let's style the quote block like the design suggests globally
- [x] The headings are forcing the font size using pixels, we should redefine the values for the font size ranges on theme.json
- [x] Let's style links globally so the links on the comments block look ok
- [ ] Spacing between paragraph blocks needs to be bigger
- [x] Spacing in between blocks in general are too tight. We might need to tweak this both globally and individually

carolinan:
- [x]  Hide Avatar in comments area

carolinan:
> - [ ] Spacing between paragraph blocks needs to be bigger

The line height for the post content is 1.6, which should match the 160% in figma, no?

carolinan:
I think that json style should be moved from the post-content block to styles > typography?

MaggieCabrera:
> > * [ ]  Spacing between paragraph blocks needs to be bigger
> 
> The line height for the post content is 1.6, which should match the 160% in figma, no?

More like margin after paragraphs in this case

PR #19, opened 2023-03-10
Add Blue Note

carolinan:
Add a rough start for Blue Note.

What's missing:
1. Font sizes, spacing (if custom)
3. Detailed templates, sidebar
4. Screenshot
5. Categories pattern for Home

Known issues:
The contrast and primary color has the same color code


blue-note/style.css
```diff
@@ -0,0 +1,16 @@
+/*
+Theme Name: Blue Note
+Theme URI: https://github.com/WordPress/community-themes/
+Author: the WordPress team
+Author URI: https://wordpress.org
+Description: 
+Requires at least: 6.1
+Tested up to: 6.2
+Requires PHP: 8.0
```
MaggieCabrera, line 9: why 8.0? are we using something specific to have that high a requirement?

blue-note/style.css
```diff
@@ -0,0 +1,16 @@
...
+Version: 0.0.1
+License: GNU General Public License v2 or later
+License URI: http://www.gnu.org/licenses/gpl-2.0.html
+Template:
```
MaggieCabrera: can this be removed?

blue-note/style.css
carolinan: No, but we should encourage users to use a newer PHP version. Consider that the current version of PHP is 8.2.

blue-note/functions.php
```diff
@@ -0,0 +1,22 @@
+<?php
+/**
+ * Register block styles.
+ * TODO: Add CSS, repeat for both featured image and image blocks
```
MaggieCabrera: I wonder if this should be a style for the query loop instead of the images and have it apply the style for the odd/even images in the loop. Are there any issues in Gutenberg about styling differently each element of the loop? historically we've done that using more than one loop and offsetting them, but that's clunky at best and a power user move. Maybe this is the moment to find a solution for this in the editor.

blue-note/style.css
carolinan:
```suggestion
```

blue-note/functions.php
carolinan: Yes I think we will run into issues. Will be interesting to see.

blue-note/templates/home.html
```diff
@@ -0,0 +1,57 @@
+<!-- wp:template-part {"slug":"header","tagName":"header"} /-->
+
+<!-- wp:group {"tagName":"main","layout":{"type":"constrained","justifyContent":"left"}} -->
+<main class="wp-block-group">
```
MaggieCabrera, line 4: I think the content of the home page should be wide width instead of full width, it looks a bit weird on very big screens

blue-note/functions.php
carolinan: There is one section that was in the middle of a page, that was like a quote with a slanted image: https://github.com/WordPress/community-themes/issues/6

blue-note/templates/home.html
```diff
@@ -0,0 +1,57 @@
...
+
+	<!-- wp:spacer -->
+	<div style="height:100px" aria-hidden="true" class="wp-block-spacer"></div>
+	<!-- /wp:spacer -->
+
+	<!-- wp:heading -->
+	<h2 class="wp-block-heading">By Topic</h2>
+	<!-- /wp:heading -->
+
+	<!-- wp:query {"queryId":0,"query":{"perPage":3,"pages":0,"offset":0,"postType":"post","order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":true},"displayLayout":{"type":"flex","columns":3},"align":"full","layout":{"type":"default"}} -->
```
MaggieCabrera: These are not really posts, but categories. How should we show these?

blue-note/theme.json
```diff
@@ -0,0 +1,224 @@
...
+				},
+				{
+					"slug": "contrast",
+					"name": "Contrast",
+					"color": "#000000"
+				},
+				{
+					"slug": "primary",
+					"name": "Primary",
+					"color": "#000000"
```
MaggieCabrera: I think we can remove the repeated color and consider the naming of the other two. This theme was designed before TT3 and was using a different convention, I believe it's best if we follow the direction of TT3 instead

blue-note/theme.json
```diff
@@ -0,0 +1,224 @@
...
+				{
+					"slug": "tertiary",
+					"name": "Tertiary",
+					"color": "#ADADAD"
+				}
+			],
+			"duotone": [
+				{
+					"slug": "duotone-1",
+					"name": "Duotone 1",
```
MaggieCabrera: I think it's more helpful for the user if these are a bit more descriptive like "black and red" for example

blue-note/theme.json
```diff
@@ -0,0 +1,224 @@
...
+					"colors": [
+						"#000000",
+						"#FFFFFF"
+					]
+				}
+			]
+		},
+		"layout": {
+			"contentSize": "655px",
+			"wideSize": "874px"
```
MaggieCabrera: these values are a bit strange, and I see looking at the figma how it's not clear what they should be. I'd settle with `contentSize` as 874px and maybe add padding to the cases where we want it narrower (or find a middle ground and equalize all of the widths) and have `wideSize` be 1300px

blue-note/templates/home.html
carolinan: Maybe, I picked the wide width from the archive page on Figma, so the wide width is much narrower than for example in TT3.
I think it is full width in the design, as it is aligns with the header and footer.

blue-note/functions.php
MaggieCabrera: I see, do you think we should have a "rotate" option for certain blocks in the editor?

MaggieCabrera:
Thanks so much for having a go at this! I will do a more thorough review of this on Monday, but it's looking like a great start!

blue-note/templates/home.html
MaggieCabrera: @beafialho if you can confirm this, it would be very helpful, thank you!

blue-note/templates/home.html
carolinan: I think the archive title and the archive description only work on the archive template.
Maybe it is an advanced variation of a category block? Untested.

blue-note/theme.json
carolinan: yes, agreed, but I did not want to make the decision alone =D

carolinan: The design only had one color per duotone, so I kept the dark color black.

blue-note/functions.php
carolinan: I'm not fully convinced that rotating images should be a block option in core or Gutenberg, it is not so common.

MaggieCabrera: https://github.com/WordPress/gutenberg/issues/40903 I think it's similar to this issue

blue-note/templates/home.html
beafialho: For the home template, the header and footer should be full width but the content can be wide.

MaggieCabrera: thank you for the feedback @beafialho and thank you for mocking it up too!

Home (or Front Page)

blue-note/functions.php
carolinan: I have added the slanted image style variation to featured images, images and the query.
But I am not sure yet how to do the duotone for the query.

I also found that the block style is not working in the site editor, but is working in the block editor and front. I think this issue is already reported, I just need to find it.

MaggieCabrera: let's get the general issues done here and we can focus on separate PRs on each template, that way is easier to review and for others to contribute too.

carolinan: I suppose we could hard code the duotone image filter...

MaggieCabrera: are you trying to have the block style do both the rotation and the duotone filter? I'm not sure I understand the problem.

And about the query loop, if you mean it for the home page, where every image has a different angle and color, we need three query loops one after the other, then the second loop will be offset by 1 post and the third by two posts.

carolinan: I believe it can be done with CSS only, without multiple queries

MaggieCabrera: mmh, sure, but then the user can only turn everything on and off, right? multiple queries are clunky but customizable via the editor. You might want to only remove the rotation or only change the duotone to better fit your palette. The only thing that the editor doesn't allow right now is the rotation, I think that's the only CSS we should add if there's a way to control the rest via the blocks themselves

carolinan: I would think anything is easier than managing 3 queries (in 3 template since the same feature are in the archive and search).
But if this has been done in .com themes before, and users have not brought it up as an issue, I guess it is a non issue.

The duotone palette can be updated in the Site Editor Styles section, right? Or is that a Gutenberg only feature?

MaggieCabrera: We've done this a a few times, yes. In the end, power users will use the list view and see the three queries and non power users will simply click on a block and expect to be able to change the duotone and be able to do so. It's not ideal but it works.

The duotone palette can't be changed yet, I'm working on a PR for that but I don't know how long it will take, there's some design of the UI that needs to happen, right now it can only be changed by modifying theme.json.

blue-note/templates/home.html
MaggieCabrera: https://github.com/WordPress/gutenberg/issues/49094 I think this is what we want

carolinan:
What is needed in order to commit this as a starting point?

MaggieCabrera (approved):
I think this is a great starting point, thank you so much @carolinan for getting this one started! Lets open issues and work on the templates and patterns on separate PRs now

blue-note/style.css
carolinan: TIL: _Create block theme_ does not support PHP 8, so this should be lowered until the plugin is updated 😞

MaggieCabrera: My dev env has wp_debug enabled and with PHP 8 I get a ton of warnings that are coming from core. I think in general going with 7 for now should be fine until core updates. In any case, CBT should update as well, I believe I saw an issue raised in their repo about it

PR #21, opened 2023-03-23
Add TT1 blocks

### LINKED ISSUE #17: Add TT1 blocks
I propose adding TT1 Blocks to this repository. I would like to see it updated.
https://wordpress.org/themes/tt1-blocks/

MaggieCabrera:
This PR adds TT1 blocks to this repo. We should continue the work started in https://github.com/WordPress/theme-experiments/pull/307 on this repo now and anything else that needs updating since the last time this theme had any work done.

/cc @carolinan 

Closes https://github.com/WordPress/community-themes/issues/17


tt1-blocks/index.php
```diff
@@ -0,0 +1,11 @@
+<?php
```
MaggieCabrera, line 1: I think this file can go

tt1-blocks/functions.php
```diff
@@ -0,0 +1,91 @@
...
+ * @return void
+ */
+function tt1_blocks_block_editor_script() {
+
+	wp_enqueue_script( 'tt1-blocks-unregister-block-style', get_theme_file_uri( '/assets/js/unregister-block-style.js' ), array( 'wp-blocks', 'wp-dom' ), wp_get_theme()->get( 'Version' ), true );
+}
+add_action( 'enqueue_block_editor_assets', 'tt1_blocks_block_editor_script' );
+
+// Block Styles.
+require get_template_directory() . '/inc/block-styles.php';
```
MaggieCabrera, line 91: I think everything in this file except the style shared/blocks stylesheets and the block styles stuff can go, don't you think @carolinan ? 

We can also discuss if those stylesheets can be consolidated into one if they end up being necessary at all

MaggieCabrera:
I think let's merge this one and continue working on separate PRs so the diffs are more meaningful to review, I think this still has a lot of work and there's a lot we can remove.

richtabor:
I don't think this theme should be part of the repo. It's quite dated/uses a lot of CSS, functions, and isn't a great example of what a top-notch block theme is (today). 

There has been a lot of developments around theming since this was originally released.

carolinan:
It wont be dated once we are ready to release the update.  🙂

carolinan:
To extend on my comment, yes it is dated right now, I am suggesting breaking backwards compatibility and updating the minimum required version to 6.2 and modernizing it.
Updating the minimum required version will have a minimal impact because: `Active Installations: 100+`

MaggieCabrera:
The reason why I think it's a good idea to include it in this repo is because TT1 blocks is already a community theme. It's already released under the WP username, and it's not being maintained. This way at the very least we ensure that the code is updated should people want to reference it at any point.

tt1-blocks/index.php
MaggieCabrera: we can do it on a follow up where we also update the min required version of WordPress for the theme

scruffian:
I think we should remove the functions.php and all the CSS files.

MaggieCabrera:
> I think we should remove the functions.php and all the CSS files.

+1, we should lower the burden of maintenance as much as we can

MaggieCabrera:
I opened issues for the removal of the code and other updates, if someone can ✅ this one, we can get the theme in the repo and start working on it?

tt1-blocks/functions.php
carolinan: I agree, and I think with more testing, many if not all the block styles should be removed too.

carolinan (approved):

MaggieCabrera:
Thank you @carolinan let's make this one pretty

PR #26, opened 2023-03-24
Stacks: update template and default styling

SaxonF:
Updates to styling , templates and pattern.


MaggieCabrera:
Thank you, Saxon, for taking the time to update the theme!

I have a couple of comments:

Can we add a link color that is more readable on darker backgrounds?

Screenshot 2023-03-24 at 09 29 00

With these changes I think we should also update the theme screenshot since it no longer represents the look and feel of the theme

stacks/patterns/home.php
```diff
@@ -3,109 +3,70 @@
  * Title: Stacks
  * Slug: stacks/home
  * Block Types: core/post-content
- * Post Types: post, page
```
scruffian, line 6: Did you mean to remove this? The idea is that this will prompt you to add the pattern when you create a new post or page.

stacks/templates/front-page.html
```diff
@@ -0,0 +1 @@
+<!-- wp:pattern {"slug":"stacks/home"} /-->
```
scruffian, line 1: Why front-page.html rather than home.html?

stacks/theme.json
```diff
@@ -41,26 +51,72 @@
...
 	"styles": {
+		"color": {
+			"background": "var:preset|color|base",
+			"text": "var:preset|color|contrast"
+		},
 		"elements": {
 			"button": {
 				"color": {
-					"background": "var(--wp--preset--color--primary)",
+					"background": "var:preset|color|contrast",
```
scruffian, line 92: I prefer the CSS syntax here, but either way I think we should be consistent.

stacks/templates/front-page.html
SaxonF: Thinking was this pattern or type of experience is better suited to single page sites, which front-page is ideal for.

stacks/patterns/home.php
SaxonF: Related to the point about single page sites, this was stopping the pattern from being loaded in home.html so the template was blank.

scruffian: I see this as a different thing. The idea of this line is that you see this when adding a block to a new post:
Screenshot 2023-03-27 at 22 00 53

SaxonF: Still shows in post editor doesn't it? I thought that attribute just dictates where that pattern is available, and by defining post/page you're stating it shouldn't be accessible in site editor. 

image

stacks/theme.json
SaxonF: theme.json was just output from create block plugin but I don't have personal preference so I'm happy to change if you'd like

stacks/patterns/home.php
scruffian: Hmm, I don't see it...

stacks/patterns/home.php
```diff
@@ -3,109 +3,70 @@
...
-<div class="wp-block-group has-white-background-color has-background" style="min-height:100vh"><!-- wp:group {"layout":{"type":"constrained"}} -->
-<div class="wp-block-group">
 
-<!-- wp:heading {"level":1,"fontSize":"x-large"} -->
-<h1 class="wp-block-heading has-x-large-font-size"><?php echo esc_html__( 'Create Stacks.', 'stacks' ); ?></h1>
+<!-- wp:group {"style":{"dimensions":{"minHeight":"100vh"},"position":{"type":"sticky","top":"0px"},"spacing":{"padding":{"top":"var:preset|spacing|70","right":"var:preset|spacing|70","bottom":"var:preset|spacing|70","left":"var:preset|spacing|70"}},"border":{"radius":"25px"}},"backgroundColor":"primary","layout":{"type":"flex","orientation":"vertical","verticalAlignment":"top","justifyContent":"left"}} -->
+<div class="wp-block-group has-primary-background-color has-background" style="border-radius:25px;min-height:100vh;padding-top:var(--wp--preset--spacing--70);padding-right:var(--wp--preset--spacing--70);padding-bottom:var(--wp--preset--spacing--70);padding-left:var(--wp--preset--spacing--70)"><!-- wp:group {"layout":{"type":"constrained"}} -->
```
scruffian, line 11: I'm not sure about this border radius - I see the white background coming through, which looks a bit strange.

MaggieCabrera (commented):
The screenshot is looking a little cut-off. I think it needs to be 1200x900

Screenshot 2023-03-30 at 15 49 35

MaggieCabrera (dismissed):
We still need to fix the link color in the pattern:

Screenshot 2023-03-30 at 15 52 20

stacks/patterns/home.php
scruffian: Ah it shows on new pages not new posts...

MaggieCabrera: Btw, if we don't chose the pattern and write regular content for a page, the frontend just shows a white screen

scruffian (approved):
I'm happy for this to be merged. I'm still not 100% on the border radius on the groups, but I wouldn't want to block on that. We can follow up on it.

PR #29, opened 2023-03-31
TT1-blocks: Update theme json

carolinan:
WIP


tt1-blocks/assets/css/blocks.css
```diff
@@ -179,23 +144,7 @@ hr.is-style-twentytwentyone-separator-thick,
 	border-bottom-width: 3px;
```
MaggieCabrera, line 140: I don't think we can remove this, can we?

tt1-blocks/assets/css/blocks.css
```diff
@@ -179,23 +144,7 @@ hr.is-style-twentytwentyone-separator-thick,
...
-	border-bottom: none;
-}
-
-.wp-block-separator.is-style-dots > hr {
-	display: none;
-}
-
 /* In Twenty Twenty-One, separators are 100% wide. This replicates that approach */
 .wp-block-separator:not(.is-style-wide):not(.is-style-dots) {
 	width: 100%;
```
MaggieCabrera, line 145: same here, it will alter people's websites

tt1-blocks/assets/css/blocks.css
```diff
@@ -67,10 +58,6 @@
 # Latest Posts
 --------------------------------------------------------------*/
 
-.wp-block-latest-posts {
-	padding-left: 0;
-}
-
 .wp-block-latest-posts.is-style-twentytwentyone-latest-posts-dividers {
```
MaggieCabrera, line 61: This sucks cause the style variation doesn't even look like the original theme. Can we get rid of it altogether and just make it into a block pattern?

MaggieCabrera:
I gave this a quick pass today, would love to hear your thoughts on most of the reminder of the CSS. I'm happy to keep minor things like the offset on text decoration or anti aliasing on the fonts, but I'd love to get rid of most of the block style stuff if at all possible but it would be disruptive. What do you think?

tt1-blocks/assets/css/blocks.css
carolinan: Yes, and it can always be adjusted in a later PR.

tt1-blocks/assets/css/blocks.css
carolinan: > Active Installations: 100+

I think breaking some backward compatibility is OK because the users still gain more features that did not exist when the theme was released....

PR #36, opened 2023-05-03
update pattern radius

SaxonF:
Updates border radius of first and last group in pattern to remove the gap.


MaggieCabrera (approved):
this looks better, thanks!

PR #39, opened 2023-06-08
TT1 Blocks: Remove parts of functions.php

### LINKED ISSUE #22: TT1 blocks: Remove unused code from functions.php
Let's clean up this file, we probably don't need most of it.

scruffian:
This removes parts of the code from functions.php, for TT1-blocks. I think we should reduce the complexity of this theme by removing block styles etc. so that its easier to maintain.

Closes #22


tt1-blocks/functions.php
```diff
@@ -1,91 +0,0 @@
...
-	 * @return void
-	 */
-	function tt1_blocks_setup() {
-		/*
-		 * Make theme available for translation.
-		 * Translations can be filed in the /languages/ directory.
-		 * If you're building a theme based on TT1 Blocks, use a find and replace
-		 * to change 'tt1-blocks' to the name of your theme in all the template files.
-		 */
-		load_theme_textdomain( 'tt1-blocks', get_template_directory() . '/languages' );
```
scruffian, line 30: This should happen automatically now.

tt1-blocks/functions.php
```diff
@@ -21,31 +21,6 @@
...
-
-		// Add default posts and comments RSS feed links to head.
-		add_theme_support( 'automatic-feed-links' );
```
scruffian, line 33: This seems to happen automattically.

tt1-blocks/functions.php
```diff
@@ -21,28 +21,6 @@
...
-		/*
-		 * Enable support for Post Thumbnails on posts and pages.
-		 *
-		 * @link https://developer.wordpress.org/themes/functionality/featured-images-post-thumbnails/
-		 */
-		add_theme_support( 'post-thumbnails' );
-		set_post_thumbnail_size( 1568, 9999 );
-
-		// Add support for Block Styles.
-		add_theme_support( 'wp-block-styles' );
```
scruffian, line 44: Let's remove block styles for this theme.

tt1-blocks/functions.php
```diff
@@ -73,19 +59,3 @@ function tt1_blocks_enqueue_styles() {
...
-/**
- * Enqueue block editor script.
- *
- * @since 0.1
- *
- * @return void
- */
-function tt1_blocks_block_editor_script() {
-
-	wp_enqueue_script( 'tt1-blocks-unregister-block-style', get_theme_file_uri( '/assets/js/unregister-block-style.js' ), array( 'wp-blocks', 'wp-dom' ), wp_get_theme()->get( 'Version' ), true );
```
scruffian, line 86: I'm not sure why wanted to do this.

MaggieCabrera: because we were forcing the width of the main separator via css and both were the same (default and wide were both 100%) and that was redundant. This needs to go

tt1-blocks/functions.php
MaggieCabrera: fine by me but we need to follow up with alternative to the block styles in the form of block patterns most likely

MaggieCabrera (approved):
Thank you, this lgtm, we should do something about the block styles on a follow up PR

PR #40, opened 2023-06-08
Blue Note: Style updates

scruffian:
This updates the colors for links and buttons so that they match the rest of the theme. Also adds rules for textInputs, but that will need this PR to work: https://github.com/WordPress/gutenberg/pull/51337


MaggieCabrera (approved):
Ship it

PR #44, opened 2023-06-08
archive template

carolinan:
username: vaibhavagarr


carolinan:
While we worked on this template, we noticed a strange bug in the Site Editor, where the page number font size did not affect the numbers that were not linked.
But, I am unable to reproduce that issue now.
I am using Chrome, on MacOS, WordPress 6.2.2, Gutenberg **trunk**.
Perhaps it is an issue that is solved on trunk but not in the latest release.

Issue #12, opened 2023-02-15
[Blue Note]: 404 template


MaggieCabrera:
![404](https://user-images.githubusercontent.com/3593343/219084224-bcb1142c-6a82-4dfb-b24b-9ed9b80c18da.jpg)

kimclowa8c:
I'll work on the 404.

kimclow:
I'll work on the 404.

kimclow:
I had to add some custom CSS to drop the background colour of the search bar: 

`.wp-block-search .wp-block-search__input {
	background: transparent;
}`

MaggieCabrera:
ah, that's a good one, we'll have to also open a Gutenberg PR so that there are controls in the future to change this, but in the meantime using CSS is fine for this

kimclow:
Update on this issue: 
I have got everything to match Figma (I think) except for the opacity of the word "search...". 

I am going to circle back to this in an hour or so after a break.

MaggieCabrera:
It would be awesome if you could open a PR with what you have, even if it's not all done. I think we can go without the opacity change in this case.

kimclow:
Updated custom CSS used for this: 


.wp-block-search .wp-block-search__input {
	background: transparent;
	height: 77px;
	padding-left: 25px;
	padding-top: 22px;
	padding-bottom: 22px;
}

kimclow:
@MaggieCabrera  - I don't actually know how to create a PR. I'll confirm once I've figured it out and submitted it :)

MaggieCabrera:
> @MaggieCabrera - I don't actually know how to create a PR. I'll confirm once I've figured it out and submitted it :)

Feel free to ask me or @carolinan for support!

kimclow:
Hey @MaggieCabrera 
I've submitted the PR. Sorry for the delay.

kimclow:
Exciting! Thank you doing the parts on this I didn't know or understand @MaggieCabrera

MaggieCabrera:
To do:

- [ ] Match the spacing between header and heading and heading and input to the design
- [ ] The typography settings for the heading are not like the design
- [ ] The input is wide width, it should be content width instead
- [ ] Add min-height to the group containing the form so we push the footer down with it

getdave:
> To do:
> 
> * [ ]  Match the spacing between header and heading and heading and input to the design
> * [ ]  The typography settings for the heading are not like the design
> * [ ]  The input is wide width, it should be content width instead
[...]

PR #41, opened 2023-06-08
Add support for textInput elements

scruffian:
This updates the approach for textInput elements to match the proposal in https://github.com/WordPress/gutenberg/pull/51337

PR #45, opened 2023-06-10
Added 404 template

kimclow:
### Changes done in this PR

- Updated 404 template to match [Figma](https://www.figma.com/community/file/1207273645711252975). 
- Added additional CSS into the style.css file. 

Mentions #12

### DIFF
--- a/blue-note/style.css
+++ b/blue-note/style.css
@@ -13,3 +13,14 @@ License URI: http://www.gnu.org/licenses/gpl-2.0.html
 Text Domain: blue-note
 Tags: custom-colors, custom-logo, custom-menu, editor-style, featured-images, blog, block-patterns, one-column, two-columns, three-columns, four-columns, rtl-language-support, threaded-comments, translation-ready, wide-blocks, full-site-editing
 */
+
+/*
+Kim Clow's suggested CSS for the 404 page. Refer to issue #12 in the repo for details.
+*/
+.error404 .wp-block-search .wp-block-search__input {
+	background: transparent;
+	height: 77px;
+	padding-left: 25px;
+	padding-top: 22px;
+	padding-bottom: 22px;
+}
--- a/blue-note/templates/404.html
+++ b/blue-note/templates/404.html
@@ -1,19 +1,3 @@
 <!-- wp:template-part {"slug":"header","tagName":"header"} /-->
-
-<!-- wp:group {"tagName":"main","layout":{"type":"constrained","justifyContent":"left","contentSize":"1320px"}} -->
-<main class="wp-block-group"><!-- wp:heading {"level":1} -->
-<h1 class="wp-block-heading">This page could not be found.</h1>
-<!-- /wp:heading -->
-
-<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"bottom":"60px","top":"59px"}}},"layout":{"type":"constrained","contentSize":"650px"}} -->
-<div class="wp-block-group alignfull" style="padding-top:59px;padding-bottom:60px"><!-- wp:paragraph {"style":{"spacing":{"padding":{"top":"0px"}},"typography":{"fontSize":"22px"}}} -->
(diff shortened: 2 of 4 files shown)


carolinan:
Thank you for working on this.

 

- [x] Minor: The title font size and font weight needs to be adjusted to match the design (80px, 400 font weight) The preferred place to add the style would be in theme.json > styles > elements > H1.

We would need to manually enqueue the stylesheet for both the editor and front. Let's try adding it in theme.json instead?

_I did not fully understand why the padding needed to be on the input field and not the block._

blue-note/style.css
```diff
@@ -12,4 +12,15 @@ License: GNU General Public License v2 or later
...
 Text Domain: blue-note
 Tags: blog, block-patterns, full-site-editing
-*/
\ No newline at end of file
+*/
+
+/*
+Kim Clow's suggested CSS for the 404 page. Refer to issue #12 in the repo for details.
+*/
+.error404 .wp-block-search .wp-block-search__input {
```
carolinan, line 20: I would recommend having the same style for the block, no matter which template it is used in?
I meant, I think we should remove `.error404`?

blue-note/style.css
```diff
@@ -12,4 +12,15 @@ License: GNU General Public License v2 or later
...
+	background: transparent;
+	height: 77px;
```
carolinan, line 22: The height here is interesting because even with https://github.com/WordPress/community-themes/pull/41/files,
we would not be able to add the height to the input field without custom CSS.

blue-note/templates/404.html
```diff
@@ -1,11 +1,19 @@
...
-<main class="wp-block-group">
+<!-- wp:group {"tagName":"main","layout":{"type":"constrained","justifyContent":"left","contentSize":"1320px"}} -->
+<main class="wp-block-group"><!-- wp:heading {"level":1} -->
+<h1 class="wp-block-heading">This page could not be found.</h1>
+<!-- /wp:heading -->
 
-<!-- wp:pattern {"slug":"blue-note/hidden-404"} /-->
+<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"bottom":"60px","top":"59px"}}},"layout":{"type":"constrained","contentSize":"650px"}} -->
+<div class="wp-block-group alignfull" style="padding-top:59px;padding-bottom:60px"><!-- wp:paragraph {"style":{"spacing":{"padding":{"top":"0px"}},"typography":{"fontSize":"22px"}}} -->
+<p style="padding-top:0px;font-size:22px">The page you were looking for could not be found. It might have been removed, renamed, or did not exist in the first place.</p>
```
carolinan: As a follow up we should either remove the pattern or move this content to the pattern.

kimclow:
Hey @carolinan 
Thanks for the feedback. 

Your minor [feedback here](https://github.com/WordPress/community-themes/pull/45#issuecomment-1588731229) has the box checked. Just confirming, did you make this change? 

> We would need to manually enqueue the stylesheet for both the editor and front. Let's try adding it in theme.json instead?

I'm afraid I'm not a developer - I work mainly in the site editor and in blocks. Can you assist with this? 

> I did not fully understand why the padding needed to be on the input field and not the block.

The padding is for the word "search" inside of the input field. It's not for the block itself. 

> I would recommend having the same style for the block, no matter which template it is used in?
> I meant, I think we should remove .error404?

This was done purposefully in case different formatting is required for other search blocks throughout the site, perhaps on different background colours. However, happy to remove the block specification if you'd prefer to. 

> The height here is interesting because even with https://github.com/WordPress/community-themes/pull/41/files,
> we would not be able to add the height to the input field without custom CSS.

Yeah, the height in Figma was specified, but there isn't an option to set the height in the site editor, so I used CSS. 

> As a follow up we should either remove the pattern or move this content to the pattern.

Okay - I'm not familiar with this process but will check it out!

carolinan:
I can make the code changes, but I wanted to discuss them first.

MaggieCabrera (approved):

PR #48, opened 2023-06-13
Blue Note: Add heading sizes and line height

carolinan:


carolinan:
Added the names to the font sizes, because I forgot them earlier
Font size options in the block settings sidebar

PR #53, opened 2023-06-15
Remove invalid tag

kafleg:
theme-options tag is only valid when Has theme options (customizer) https://make.wordpress.org/themes/handbook/review/required/theme-tags/


MaggieCabrera (approved):
Thank you!

PR #59, opened 2023-07-20
Update PHP min version required for themes

MaggieCabrera:
Let's update the themes to require the same [PHP version that WordPress does](https://wordpress.org/about/requirements/)


carolinan (approved):

Soean:
WordPress 6.3 requires PHP 7.0, it just recommend PHP 7.4.

Issue #5, opened 2023-02-15
[Blue Note]: Home template


MaggieCabrera:
**Desktop:**

![Home (or Front Page)](https://user-images.githubusercontent.com/3593343/219079212-70718225-8ee4-4c37-b78c-5a2a238fd225.jpg)

**Mobile**

![Homepage - Mobile](https://user-images.githubusercontent.com/3593343/219079349-ae2ac68b-2ac3-4c75-aea9-d10c69c87376.jpg)

beafialho:
@MaggieCabrera one alternative for the current lack of option to display categories with images, is to just remove that "By Topic" section and leave only the posts.

MaggieCabrera:
> @MaggieCabrera one alternative for the current lack of option to display categories with images, is to just remove that "By Topic" section and leave only the posts.

Yes, that's an idea. I would like to explore the possibility of adding the functionality in the editor, since it's a common pattern for themes, but if it's a blocker, we can totally remove it. Thanks!

beafialho:
Cool, whatever you think is best works for me, I was only leaving that option on the table 👍🏼

carolinan:
Where does the button "see posts" lead to?

MaggieCabrera:
> Where does the button "see posts" lead to?

Let's go with the archive page

Mooore:
I'll try this :)

MaggieCabrera:
I added you to the repo!

MaggieCabrera:
I remember someone working on this one, but I don't see any changes to the code. Do you have a zip for this, @carolinan?

MaggieCabrera:
I found it! After https://github.com/WordPress/community-themes/pull/61 is merged, what's left to do here is:

- [ ] Add duotone to the images
- [ ] Improve spacing between the query loop and the hero section
- [ ] Remove underline from post titles
- [ ] Font sizes are not correct
- [ ] Reduce line height from post titles
- [ ] Improve spacing between posts vertically
- [ ] The line height on the footer is also too big
- [ ] The social menu should be aligned vertically to the text on the left

carolinan:
> > Where does the button "see posts" lead to?
> 
> Let's go with the archive page

So `page/2/` ?

carolinan:
If we could customize the query pagination text like we can with the post navigation link text, that would solve this problem.

PR #61, opened 2023-07-20
Blue note: Home template

MaggieCabrera:
I had the zip file for the home page in my DMs! Thank you @Mooore for working on this


scruffian:
I made a change to the footer as I think this is the design intent:
Screenshot 2023-07-24 at 10 30 07

MaggieCabrera:
Looking at the design I think all of the content in the black area should extend the same width, and probably make it wide width, I'm not sure what it is from your screenshot.

MaggieCabrera:
I made comments on https://github.com/WordPress/community-themes/issues/5 but yeah, aligning the footer with the content should be on that list

scruffian:
Like this?
Screenshot 2023-07-24 at 14 03 19

MaggieCabrera:
yeah, it makes sense to me that either everything align or none of it is

carolinan:
The no results block only contains the placeholder, compared to https://github.com/WordPress/community-themes/pull/65/files#diff-be779912416c3055199e2a71598bfa1e684f93750edd3769fafd203ac2079989R35

scruffian (approved):
This is looking good. I think we should merge and iterate.

PR #67, opened 2023-07-25
Update single with sidebar + large header template part + sidebar

carolinan:
- [x] Post title and date should be centered vertically
- [x] The headings are forcing the font size using pixels, we should redefine the values for the font size ranges on theme.json
- [x]  Spacing in between blocks in general are too tight. We might need to tweak this both globally and individually
- [x] Hide Avatar in comments area
- [x]  Use the post meta pattern
- [x]  Make the sidebar column use `<aside>`
**To do:** 
- [x]  After adding the wrapping group for the aside, the block spacing between the sections in the sidebar is off again.


MaggieCabrera:
Oh, I forgot to add to the list, can we change the container for the sidebar to use the `aside` tag?

carolinan:
Columns can not have a different HTML element but we can wrap it in a group.

blue-note/templates/single.html
```diff
@@ -1,27 +1,32 @@
...
+
+<!-- wp:group {"style":{"spacing":{"blockGap":"var:preset|spacing|20"}},"layout":{"type":"flex","flexWrap":"nowrap"}} -->
+<div class="wp-block-group">
+	<!-- wp:paragraph -->
+	<p>Author: </p>
+	<!-- /wp:paragraph -->
+	
+	<!-- wp:post-author-name /-->
+	
+	<!-- wp:paragraph -->
```
MaggieCabrera: Can we use [this pattern](https://github.com/WordPress/community-themes/blob/trunk/blue-note/patterns/post-meta-nav.php) for this bit?

blue-note/theme.json
```diff
@@ -223,6 +223,48 @@
 					}
 				}
 			},
+			"core/comment-content": {
+				"css": " & a{text-decoration:underline;} & a:hover{text-decoration:none;}"
```
carolinan, line 227: Enforces underlined links in comment content

blue-note/theme.json
```diff
@@ -223,6 +223,48 @@
...
+			},
+			"core/comment-template": {
+				"elements": {
+					"link": {
+						"typography": {
+							"textDecoration": "none"
```
carolinan, line 233: Removes the underline from the default state of the links elsewhere in the comment area.
Adds the underline on :hover.

**Note:** I did not change the text color because the suggested color has too low color contrast against the background color.

scruffian:
> After adding the wrapping group for the aside, the block spacing between the sections in the sidebar is off again.

I pushed a commit to fix this.

scruffian (approved):
This looks good, let's bring it in.

Issue #11, opened 2023-02-15
[Blue Note]: Search template


MaggieCabrera:
![Search](https://user-images.githubusercontent.com/3593343/219084141-a8702491-cf9f-4bb8-8b70-0a3184d67e63.jpg)

MaggieCabrera:
To do:

- [ ] Search results are not centered
- [ ] Search results need to be styled like the design. We can probably use a pattern for this, since the archive template uses a similar style
- [ ] Search input could use the same CSS that is applied to the 404 form input, just needs adding the search template specific classes
- [ ] Match the spacing between the heading and the header
- [ ] Match the heading typography settings to the design
- [ ] The pagination could go without the numbers
- [ ] Let's limit the loop to 3 posts

getdave:
I'll take a look at these.

I'll update as I go...

>  Search results are not centered

✅ These seem to be centered to me.

> Search results need to be styled like the design. We can probably use a pattern for this, since the archive template uses a similar style

✅ Already using `<!-- wp:pattern {"slug":"blue-note/posts"} /-->`.

> Search input could use the same CSS that is applied to the 404 form input, just needs adding the search template specific classes

Let's address this via https://github.com/WordPress/community-themes/pull/71

> Match the spacing between the heading and the header

Let's try https://github.com/WordPress/community-themes/pull/73

> Match the heading typography settings to the design

✅ Seems ok to me.

> The pagination could go without the numbers

✅ Already fixed

> Let's limit the loop to 3 posts

The loop inherits from the Template query. As that comes from Site Settings we can't control that unless we use a custom query loop setting here. Is that what we want @MaggieCabrera?

PR #71, opened 2023-07-25
Utilise Pattern for search form

getdave:
This PR unifies all instances of the search form into a single Pattern.

Some instances requires wrapping in a Group block to provide the original 75% constrained width.

Side note/rant: it's really not at all intuitive to set this width in the UI on the Group block.

Part of work towards https://github.com/WordPress/community-themes/issues/11

### DIFF
--- a/blue-note/patterns/hidden-404.php
+++ b/blue-note/patterns/hidden-404.php
@@ -7,16 +7,18 @@
 ?>
 <!-- wp:group {"tagName":"main","layout":{"type":"constrained","justifyContent":"left","contentSize":"1320px"}} -->
 <main class="wp-block-group"><!-- wp:heading {"level":1} -->
-<h1 class="wp-block-heading"><?php esc_html_e( 'This page could not be found','blue-note' ); ?></h1>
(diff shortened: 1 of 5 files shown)


carolinan:
What is the benefit of creating a pattern with one single block?
(Theme patterns are not synced)

getdave:
> What is the benefit of creating a pattern with one single block? (Theme patterns are not synced)

I saw it as a maintenance things for developers. When I used to make classic themes I would have the search form as a template include. This is basically the same thing...apart from the syncing piece you mentioned.

I'm not particularly wedded to this approach so if it's not "the way things are done" then I'm happy to be advised as such 🙇

carolinan:
If we want it to sync it could be a template part, but template parts does not have an "inserter no" option.

The benefit of a pattern is that text can be translated, but this pattern does not use translation functions around the Search label.

MaggieCabrera:
Generally we try to avoid doing things just because it's more comfortable for devs if they may confuse the users. Like Carolina said, the pattern will be visible for them and don't make much sense on its own

getdave:
Cool. Let's close this then 👍

PR #72, opened 2023-07-25
Blue Note: Increase quote block default size and create example pattern 

carolinan:
Adds a block pattern for a quote with a slanted image with a duotone filter.

Quote in post with sidebar


MaggieCabrera (approved):
Thank you for making the pattern! I swapped the image to be the one from the design, which is the one that makes sense with the text

PR #73, opened 2023-07-25
Standardise spacing between Header and Main Content across all templates

getdave:
Many Issues mention spacing between the Header and the Main Content (often referred to as "the Heading" in the Issues).

Instead of correcting on a per template basis let's just set a bottom margin on the Group _within_ the Header part.

As this is a vertical spacing we should base on the viewport width so that it scales appropriately.

Ideally we'd be able to clamp this to a min-max but I don't think that's possible...it is?

Part of https://github.com/WordPress/community-themes/issues/11


scruffian (commented):
This makes sense but I think we should use one of the presets

richtabor:
> This makes sense but I think we should use one of the presets

You could set min/max with spacing presets like this: https://github.com/richtabor/wabi/blob/main/theme.json#L183

scruffian:
Tried that here: https://github.com/WordPress/community-themes/pull/77

getdave:
We should merge Ben's PR and then rebase this to use that.

scruffian (approved):
LGTM

PR #74, opened 2023-07-25
Increase spacing between header and Latest Posts section in Home Pattern

getdave:
THe spacing between the Home Pattern's Header `Learn All about Jazz` and the "Latest Posts" sectino doesn't match the design.

This PR adds spacing.

### DIFF
--- a/blue-note/patterns/home.php
+++ b/blue-note/patterns/home.php
@@ -7,8 +7,8 @@
  */
 ?>
 
-<!-- wp:group {"tagName":"main","style":{"spacing":{"margin":{"top":"100px"}}},"layout":{"type":"constrained","justifyContent":"center"}} -->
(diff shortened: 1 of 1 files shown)


scruffian (approved):
This looks good. My instinct would be to also replace the 100px with a preset too, but I'm curious what @MaggieCabrera thinks...

getdave:
> This looks good. My instinct would be to also replace the 100px with a preset too, but I'm curious what @MaggieCabrera thinks...

It should be a present yes, but can presets be fluid? Do they scale with viewport? Everything seems so static/fixed which doesn't seem great.

MaggieCabrera:
Default core presets are static I think, but they can be replaced in theme.json to use clamp or more dynamic values if you want!

PR #77, opened 2023-07-26
Blue Note: Move to fluid spacing

scruffian:
This updates Blue Note to use fluid spacing. 

These presets are slightly larger at smaller screen sizes, so I have downsized all of the presets too, to keep the spacing about the same. It's not an exact science though!

### DIFF
--- a/blue-note/theme.json
+++ b/blue-note/theme.json
@@ -504,9 +504,9 @@
 		"spacing": {
 			"padding": {
 				"top": "0",
-				"right": "var(--wp--preset--spacing--50)",
+				"right": "var(--wp--preset--spacing--60)",
 				"bottom": "0",
-				"left": "var(--wp--preset--spacing--50)"
+				"left": "var(--wp--preset--spacing--60)"
 			}
 		},
 		"typography": {
--- a/blue-note/theme.json
+++ b/blue-note/theme.json
@@ -106,6 +106,61 @@
 				"rem",
 				"vh",
 				"vw"
+			],
+			"spacingScale": {
+				"steps": 10
+			},
+			"spacingSizes": [
+				{
+					"size": "min(20px, 1vw)",
+					"slug": "10",
+					"name": "1"
+				},
+				{
+					"size": "min(30px, 2vw)",
+					"slug": "20",
+					"name": "2"
+				},
+				{
+					"size": "min(40px, 4vw)",
+					"slug": "30",
+					"name": "3"
+				},
+				{
+					"size": "min(50px, 6vw)",
+					"slug": "40",
+					"name": "4"
+				},
+				{
+					"size": "min(60px, 10vw)",
+					"slug": "50",
+					"name": "5"
+				},
+				{
+					"size": "min(70px, 12vw)",
+					"slug": "60",
+					"name": "6"
+				},
+				{
+					"size": "min(80px, 14vw)",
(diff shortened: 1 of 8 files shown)


getdave:
This feels like exactly how we should be building Themes. Spacing, especially vertical spacing, should adapt to viewport up to a max setting. It should "just work" at all sizes no matter which size you "design" the Theme on.

Thank you 👏 

I'll defer to more experienced theme designers for review.

blue-note/theme.json
richtabor, line 1: I suggest the spacing steps are maxed out at `7`, so that the standard spacing stepped RangeControl is used. Otherwise, you end up with a select control—which degrades the experience quite a bit. 

CleanShot 2023-07-26 at 14 53 32
CleanShot 2023-07-26 at 14 55 20

blue-note/theme.json
```diff
@@ -464,9 +519,9 @@
 		"spacing": {
 			"padding": {
 				"top": "0",
-				"right": "var(--wp--preset--spacing--60)",
+				"right": "var(--wp--preset--spacing--50)",
```
richtabor: I'd suggest going with 40 left and right here.

richtabor (commented):
A couple suggestion. I'd strongly consider reducing the amount of sizes uses to get them at a max of 7.

blue-note/theme.json
carolinan: The design uses a minimum of 61px on desktop.

scruffian:
I decided to just make the changes to the presets in the theme json. The slugs now match more clearly to pixel values, e.g.:

`--wp--preset--spacing--60` is `60px`

blue-note/theme.json
richtabor: Spacing size 40, not pixels. And I’d personally tune it to 60px, instead of 61.

richtabor: The current mobile width is a bit too wide as-is.

carolinan: Yes,  I understood that you meant the spacing preset =D

Spacing size 40 is a minimum of 40px until 6vw is smaller, no?

```

				{
					"size": "min(40px, 6vw)",
					"slug": "40",
					"name": "4"
				},

				{
					"size": "min(60px, 12vw)",
					"slug": "60",
					"name": "6"
				},
```

The content is not centered on the mobile designs, and some are 11 px on the left and 20 on the right. So I assume we have some freedom to make a decision and not follow the exact design.
I don't get a chance to work with a design that someone else has created very often, so I am always weary about how much I can change.

MaggieCabrera (approved):
I agree with this approach, I just wonder if we should tweak the `vw` numbers. I was mainly looking at the spacings for the header and they seem to vary very little when looking at really small viewports, so it's barely noticeable that the spacing is fluid at all.

richtabor:
> I agree with this approach, I just wonder if we should tweak the `vw` numbers. I was mainly looking at the spacings for the header and they seem to vary very little when looking at really small viewports, so it's barely noticeable that the spacing is fluid at all.

Yea they should be improved (and consolidated a bit). I don’t think 9 and 10 are used at all, you’d just to combine two spacing values that are close and then you’d have the proper spacing RangeControl.

MaggieCabrera (approved):
I think we can bring this one in and iterate on the dynamic values if we feel like we need to

Issue #78, opened 2023-07-26
Blue note: patterns


MaggieCabrera:
https://www.figma.com/file/5QJUr6hqRUClr0lz0UEsfz/Blue-Note-(Community)?node-id=35%3A254&mode=dev

The design provides a set of patterns we can include to the theme

PR #79, opened 2023-07-26
Blue note: Add patterns thin stripes and thick stripes

MaggieCabrera:
Adds the Thin Stripes pattern and a block style associated with it. Users can turn on or off the clip-path and duotone filter. If the theme changes, the image will display without changes.

Screenshot 2023-07-26 at 17 22 08

Part of https://github.com/WordPress/community-themes/issues/78


blue-note/assets/images/thin-stripes.svg
```diff
@@ -0,0 +1,31 @@
+<svg width="1322" height="1044" viewBox="0 0 1322 1044" fill="none" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
ajlende: I would remove the `width` and `height` parameters since the size should be defined by the image block and the `viewBox` _should_ (I think) be enough to position things relatively in the `clipPath` definition.

`fill` (here and below) can also be removed since we're just providing definitions.

And `xlink` can probably be removed too since we aren't using any of its features.

MaggieCabrera: lovely, I'll get to that, thank you

MaggieCabrera:
Since the second pattern is so similar, I added it to this same PR:

Screenshot 2023-07-26 at 18 11 08


@ajlende the image for this second pattern is one we are using for another pattern, with a different aspect ratio. I've played with the new aspect ratio tools to make it work with the more wide design and it seems to work in the editor (though it would be lovely to control where it cuts out) but not in the frontend for me, not sure if it's a bug you are aware of. I might end up duplicating this image and croping it for this theme, but I think it was worth mentioning anyway

scruffian:
I think we need to add a width attribute so that the aspect ratio works, which I have done.

blue-note/patterns/image-thick-stripes.php
```diff
@@ -0,0 +1,20 @@
+<?php
+/**
+ * Title: Striped image, text and button
+ * Slug: blue-note/thick-stripes
+ * Categories: text button
+ */
+?>
+<!-- wp:image {"width":847,"aspectRatio":"16/9","scale":"cover","sizeSlug":"full","linkDestination":"none","style":{"color":{"duotone":"var:preset|duotone|black-orange"}},"className":"is-style-blue-note-thick-stripes"} -->
+<figure class="wp-block-image size-full is-resized is-style-blue-note-thick-stripes"><img src="http://wp-src.test/wp-content/themes/community-themes/blue-note/assets/images/Cab_Calloway_Gottlieb.jpg" alt="" style="aspect-ratio:16/9;object-fit:cover;width:847px" width="847"/></figure>
```
carolinan: The image link needs to be updated.

scruffian (approved):
Let's bring this in...

PR #80, opened 2023-07-27
Blue note: add pattern media text

MaggieCabrera:
Adds the following pattern:

Screenshot 2023-07-27 at 10 09 51


scruffian:
I wonder if it would be better to just use columns for this?

MaggieCabrera:
> I wonder if it would be better to just use columns for this?

With columns you can't make the image adapt to the height of the other column sadly

jeryj:
Will the frontend content area width get increased? This template looks great in the editor but is really squished on the frontend.



And at smaller sizes, can it break to a top/bottom image sooner?

S

MaggieCabrera:
The pattern was designed to work better on wide width with the amount of content that it has. The user can adapt it with their own content. It will look ok if used on a page template or a post with the single no sidebar template, but it will look bad on a post that shows a sidebar if left as it is. 

The breakpoint for the block is not something we can control, and something I hesitate to change via CSS since it's specific to this pattern in particular. I'm happy to not include this one in the theme if we don't see that we can make it work with what we have at the moment.

jeryj:
That makes a lot of sense. I think it's a good pattern to preserve to demo to people what's possible. How about reducing the content to one or two shorter paragraphs and removing the left padding on the text? That would allow it to look better at most screen sizes, and at mobile view, not have a large left side padding on the text.

Admittedly, it does look really good in the editor as is, but reducing the content and removing the left padding will make it look better on the frontend and at the other screen sizes.

blue-note/patterns/media-text.php
```diff
@@ -0,0 +1,27 @@
...
+<!-- /wp:paragraph -->
+
+<!-- wp:paragraph -->
+<p><?php esc_html_e( 'However, Coltrane had yet to reach his full potential, suffering from addiction to alcohol and heroin for most of his life after serving in the Navy. Due to Coltrane’s violation of Davis’ rule barring the use of alcohol and narcotics, he was kicked out of the quintet. Determined to kick this habit, Coltrane went cold turkey, completely stopping his use of drugs, a decision which almost killed him. After moving past his addiction, Coltrane rejoined Miles Davis in a sextet, which, over two years, released several critically acclaimed records.', 'blue-note' ); ?></p>
+<!-- /wp:paragraph -->
+
+<!-- wp:paragraph -->
+<p><?php esc_html_e( 'Then, Coltrane left to lead his own band, releasing the album, Giant Steps. The album, which was entirely composed by Coltrane, was revolutionary, and within it, Coltrane used a technique for chord progressions which had never been used before, hence it being named a “Coltrane change.”', 'blue-note' ); ?></p>
+<!-- /wp:paragraph --></div>
```
scruffian, line 25:
```suggestion
</div>
```

kafleg:
@MaggieCabrera 
Tested this pattern when there is no sidebar as well, the image is cropped. But, if we disable, **Crop image to fill entire column** the image is fine. 

the below screenshot when it is disabled, 
image

draganescu (approved):
I tested:

- the pattern alone
- the pattern in a group with various content widths
- mobile and tablet rendering

I think it works great. Cropping to fill the column is nice if you set the right focal point.

MaggieCabrera:
Let's bring this one in as is, and let the users adapt it depending on how they wanna use it

PR #83, opened 2023-07-27
Blue Note: Simplify the archive template

scruffian:
This simplifies some of the layout settings in Blue Note to better match the designs

Before:
Screenshot 2023-07-27 at 14 10 46

After:
Screenshot 2023-07-27 at 14 10 25

I also removed the archive template as it will fall back to index which I think is the same.


carolinan (approved):
Won't deleting it make it more difficult for users to edit their archives? If this is something you commonly do in the block themes, and there aren't any support questions about it, then I 
don't mind at all.

scruffian:
Users can edit the index template to edit the archives. I don't think the index template is used for anything else in this theme.

PR #84, opened 2023-07-27
Blue Note: Blue Quote pattern

jeryj:
My first pattern! Feedback very welcome, as I don't know the standards here.

I took [the design from figma](https://www.figma.com/file/5QJUr6hqRUClr0lz0UEsfz/Blue-Note-(Community)?type=design&node-id=35-305&mode=design&t=XkWWU8KmKkV8uSMU-0) and recreated it as best I could using only the block editor, then added block styles to get it closer to the design.

There are some differences between the editor and frontend though. Mainly, the `cite` element in the editor adds some inline styles that are not present on the frontend. I'll look into why that is within Gutenberg, as I think that's where that change would likely be made.


blue-note/patterns/blue-image-with-quote.php
```diff
@@ -0,0 +1,25 @@
...
+
+<!-- wp:column {"verticalAlignment":"center","width":"66.6%","style":{"spacing":{"padding":{"right":"0","left":"0"},"blockGap":"0"}},"layout":{"type":"default"}} -->
+<div class="wp-block-column is-vertically-aligned-center" style="padding-right:0;padding-left:0;flex-basis:66.6%"><!-- wp:quote {"style":{"elements":{"link":{"color":{"text":"var:preset|color|white"}}},"typography":{"textTransform":"uppercase","fontStyle":"normal","fontWeight":"300"}},"textColor":"white","className":"is-style-blue-note-blue-quote","fontSize":"large"} -->
+<blockquote class="wp-block-quote is-style-blue-note-blue-quote has-white-color has-text-color has-link-color has-large-font-size" style="font-style:normal;font-weight:300;text-transform:uppercase"><!-- wp:paragraph {"style":{"color":{"text":"#4b99f8"},"spacing":{"margin":{"top":"0","bottom":"0","left":"0","right":"0"},"padding":{"right":"0","left":"0"}}}} -->
+<p class="has-text-color" style="color:#4b99f8;margin-top:0;margin-right:0;margin-bottom:0;margin-left:0;padding-right:0;padding-left:0">“Every time I close my eyes blowing that trumpet of mine—I look right in the heart of good old New Orleans… It has given me something to live for"</p>
```
MaggieCabrera: We need to make the string translateable, the 404 pattern is good example of how

blue-note/patterns/blue-image-with-quote.php
```diff
@@ -0,0 +1,25 @@
...
+<div class="wp-block-group is-style-blue-note-blue-quote has-contrast-background-color has-background" style="padding-top:var(--wp--preset--spacing--60);padding-right:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--60);padding-left:var(--wp--preset--spacing--60)"><!-- wp:columns {"verticalAlignment":"center","style":{"spacing":{"blockGap":{"left":"var:preset|spacing|70"}}}} -->
+<div class="wp-block-columns are-vertically-aligned-center"><!-- wp:column {"verticalAlignment":"center","width":"33.3%"} -->
+<div class="wp-block-column is-vertically-aligned-center" style="flex-basis:33.3%"><!-- wp:image {"id":10,"sizeSlug":"full","linkDestination":"none","style":{"color":{"duotone":["#000000","rgb(75, 153, 248)"]},"border":{"color":"#4b99f8","width":"2px"}}} -->
+<figure class="wp-block-image size-full has-custom-border"><img src="<?php echo esc_url( get_stylesheet_directory_uri() ); ?>/assets/images/Louis_Armstrong.jpg" alt="" class="has-border-color wp-image-10" style="border-color:#4b99f8;border-width:2px"/></figure>
```
MaggieCabrera: Generally we strip away the IDs from query and image blocks because those are specific to your own installation and will probably be taken in some other user's install

blue-note/patterns/blue-image-with-quote.php
```diff
@@ -0,0 +1,25 @@
+<?php
+/**
+ * Title: Blue Image with Quote
+ * Slug: blue-note/blue-quote
+ * Categories: image text
```
MaggieCabrera: I think there's no image category

jeryj:
Thanks for all the feedback and direction, @MaggieCabrera! I think I have it all addressed.

MaggieCabrera (approved):

MaggieCabrera:
Thank you @jeryj !

PR #85, opened 2023-07-27
Blue Note: Call to action pattern

jeryj:
[Pattern 13 from the Blue Note Figma](https://www.figma.com/file/5QJUr6hqRUClr0lz0UEsfz/Blue-Note-(Community)?type=design&node-id=35-254&mode=design&t=pHDOhjRNupHT8nI1-0).

I changed the copy (with the help of ChatGPT) to make it more interesting and added a paragraph. I felt the layout was a bit more realistic and playful that way.

### Frontend view
Screenshot 2023-07-27 at 3 40 04 PM


MaggieCabrera (approved):
Thank you! I'm bringing this in. I made the strings translatable and removed the block types. Those are more meant for cases like the query loop where you want the pattern to show up as an option when you select not to start blank. I don't think they serve any purpose for blocks like paragraph or group

PR #88, opened 2023-07-28
Blue note: Search form styles

### LINKED ISSUE #11: [Blue Note]: Search template

MaggieCabrera:
I think this is the only thing missing to close https://github.com/WordPress/community-themes/issues/11

### DIFF
--- a/blue-note/patterns/posts.php
+++ b/blue-note/patterns/posts.php
@@ -12,9 +12,9 @@
 	<!-- wp:post-template {"layout":{"type":"constrained"}} -->
 		<!-- wp:group {"className":"blue-note-stacked-featured-images"} -->
 		<div class="wp-block-group blue-note-stacked-featured-images">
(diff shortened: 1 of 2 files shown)


scruffian:
We should probably look at the no results variation too:

Screenshot 2023-08-02 at 11 40 53

I also find the left aligned search field very strange.

MaggieCabrera:
I don't hate it if the heading is to the left too. It's how it was on figma. The different size does bug me though.

scruffian:
The featured image is sometimes too small as well...

Screenshot 2023-08-02 at 11 48 28

MaggieCabrera:
> The featured image is sometimes too small as well...

Do you mean it should have the aspect ratio applied?

MaggieCabrera:
I think we should do the aspect ratio thing anyway since the featured image is not the same size in the three places that it shows: the home template, the archive/search and the single template. We should also add aspect ratio to the header-large pattern now that I think about it.

MaggieCabrera:
Screenshot 2023-08-03 at 10 04 32

I don't think we will see many cases where index/archive will show the no results text, so I changed it so it doesn't show a second h2 and a second search block. Also added the aspect ratio.

scruffian (approved):
LGTM. I still find the left aligned search form weird but that's the design 🤷

PR #99, opened 2023-08-11
Blue note: fix the nav block for header large template part

MaggieCabrera:
- fix markup for nav block on header large


kafleg:
What does this mean? 
image

MaggieCabrera:
that means that you have a page without a title, I believe.

kafleg (approved):

PR #113, opened 2023-12-21
Atlas: Added block styles for angled versions of starscape and cover block

MaggieCabrera:
Adds block styles for atlas to create an angled shape on cover blocks and starscape blocks

Screenshot 2023-12-21 at 19 00 00


ajlende (approved):
Looks good to me! 👍

PR #116, opened 2023-12-22
Atlas: Add home page template

scruffian:
Adds a home page template:

Screenshot 2023-12-22 at 12 45 47

It's a shame there's no way to make the cover block itself link.


atlas/templates/home.html
```diff
@@ -0,0 +1,43 @@
...
+<h1 class="wp-block-heading has-text-align-center">Atlas</h1>
+<!-- /wp:heading -->
+
+<!-- wp:heading {"textAlign":"center","level":3,"placeholder":"Write title…"} -->
+<h3 class="wp-block-heading has-text-align-center">A starscaped block theme.</h3>
```
MaggieCabrera, line 9: if we go with this we'll have to make it all translatable, but we can do that later

kafleg:
The theme name Atlas is already taken. https://themes.svn.wordpress.org/atlas/ 

We will change the name later?

MaggieCabrera:
> The theme name Atlas is already taken. https://themes.svn.wordpress.org/atlas/
> 
> We will change the name later?

good call, @ajlende do you have more ideas? @scruffian mentioned that it would be nice to try and get a community theme for each letter of the alphabet

PR #118, opened 2023-12-22
Atlas: Use hover state in the navigation

scruffian:
This adds a nice hover state to the navigation:

Screenshot 2023-12-22 at 14 02 08

It required some new spacing settings.

### DIFF
--- a/atlas/theme.json
+++ b/atlas/theme.json
@@ -83,6 +83,43 @@
 				}
 			]
 		},
+		"spacing": {
+			"spacingScale": {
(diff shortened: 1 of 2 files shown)


atlas/theme.json
```diff
@@ -115,7 +152,7 @@
 	},
 	"styles": {
 		"blocks": {
-			"core/navigation-link": {
```
scruffian, line 118: I just removed this but the diff is weird because GH is trying to be too clever.

PR #120, opened 2023-12-22
Update Atlas 404 and Single

scruffian:
This updates the 404 to this:

Screenshot 2023-12-22 at 15 14 30

And the single template to:

Screenshot 2023-12-22 at 18 43 58


atlas/templates/single.html
```diff
@@ -0,0 +1,29 @@
+<!-- wp:template-part {"slug":"header","tagName":"header"} /-->
```
MaggieCabrera, line 6: I think this shouldn't be here

atlas/theme.json
```diff
@@ -285,19 +299,14 @@
 				"typography": {
 					"fontFamily": "var(--wp--preset--font-family--dm-sans)"
 				}
-			},
-			"link": {
-				"color": {
-					"text": "var(--wp--preset--color--accent)"
```
MaggieCabrera: is this a merge conflict?

MaggieCabrera:
On Firefox I see a space over the header and there's an unneeded vertical scroll. Should we center the content vertically too?
Screenshot 2023-12-22 at 17 28 51

scruffian:
> Should we center the content vertically too?

I don't think we can without centering the header

MaggieCabrera:
> > Should we center the content vertically too?
> 
> I don't think we can without centering the header

Sounds like a challenge!

scruffian:
i have removed the padding. The vertical scroll will need custom CSS so I'd rather wait to see if we can generalise it on all templates.

MaggieCabrera:
I'm gonna look into the centering, but in the site editor I see a block error and when I resolve it it gives me a whole different markup for the stars?

PR #123, opened 2024-02-05
Update Atlas

ajlende:
WIP


MaggieCabrera:
Is it ok to work on this branch for things like translating strings or moving bits of the templates into patterns?

ajlende:
A lot of the text is copy from my website that we're going to want to replace with something else, but my work on this branch is pretty inconsistent because of other things I have going on, so feel free to commit directly to it.

Issue #135, opened 2024-02-20
Poetry: Screenshot


MaggieCabrera:
Add a proper screenshot to the theme

PR #139, opened 2024-03-02
Fix the proper screenshot image issue in Poetry theme

sandroschutt:
I have created a dummy navigation for the theme, took a screeshot and edited the image to fit the recommended requirements.


thetwopct:
Referencing the issue - https://github.com/WordPress/community-themes/issues/135

MaggieCabrera (approved):
This looks like a good starting point, we can replace later if it ends up looking any different

PR #140, opened 2024-03-02
Fix the missing archive template in Poetry

### LINKED ISSUE #131: Poetry: archive template
The theme only designs an index template, we need a template for the archive.

sandroschutt:
Creates a basic template for categories, tags and other base archives in the Poetry theme.

### DIFF
--- a/poetry/patterns/header.php
+++ b/poetry/patterns/header.php
@@ -9,16 +9,16 @@
 <!-- wp:group {"style":{"spacing":{"padding":{"top":"32px","bottom":"64px"}}},"layout":{"type":"constrained"}} -->
 <div class="wp-block-group" style="padding-top:32px;padding-bottom:64px"><!-- wp:group {"style":{"spacing":{"blockGap":"64px"}},"layout":{"type":"flex","flexWrap":"nowrap"}} -->
 <div class="wp-block-group"><!-- wp:group {"style":{"spacing":{"blockGap":"88px"}},"layout":{"type":"flex","orientation":"vertical"}} -->
-<div class="wp-block-group"><!-- wp:site-title {"level":0,"isLink":false,"style":{"typography":{"fontSize":"24px","fontStyle":"normal","fontWeight":"600"}},"textColor":"custom-text"} /-->
+<div class="wp-block-group"><!-- wp:site-title {"level":0,"isLink":false,"style":{"typography":{"fontStyle":"normal","fontWeight":"600"}},"fontSize":"medium"} /-->
 
 <!-- wp:group {"style":{"spacing":{"blockGap":"48px"}},"layout":{"type":"flex","orientation":"vertical"}} -->
-<div class="wp-block-group"><!-- wp:heading {"level":1} -->
-<h1>Food <br>for thoughts</h1>
-<!-- /wp:heading -->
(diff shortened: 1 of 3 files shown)


poetry/templates/archive.html
```diff
@@ -0,0 +1,23 @@
+<!-- wp:template-part {"slug":"header","area":"header"} /-->
+
+<!-- wp:group {"layout":{"type":"constrained"}} -->
+<div class="wp-block-group"><!-- wp:query-title {"type":"archive","style":{"spacing":{"padding":{"bottom":"var:preset|spacing|40"}}}} /-->
+
+<!-- wp:query {"queryId":32,"query":{"perPage":10,"pages":0,"offset":0,"postType":"post","order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":true},"layout":{"type":"default"}} -->
```
MaggieCabrera: We don't want the Id on query loops (explained [here](https://github.com/WordPress/community-themes?tab=readme-ov-file#pattern-creation-guidelines) under _Other tips_).

thetwopct:
Closes https://github.com/WordPress/community-themes/issues/131

poetry/patterns/header.php
```diff
@@ -9,16 +9,16 @@
 <!-- wp:group {"style":{"spacing":{"padding":{"top":"32px","bottom":"64px"}}},"layout":{"type":"constrained"}} -->
 <div class="wp-block-group" style="padding-top:32px;padding-bottom:64px"><!-- wp:group {"style":{"spacing":{"blockGap":"64px"}},"layout":{"type":"flex","flexWrap":"nowrap"}} -->
 <div class="wp-block-group"><!-- wp:group {"style":{"spacing":{"blockGap":"88px"}},"layout":{"type":"flex","orientation":"vertical"}} -->
-<div class="wp-block-group"><!-- wp:site-title {"level":0,"isLink":false,"style":{"typography":{"fontSize":"24px","fontStyle":"normal","fontWeight":"600"}},"textColor":"custom-text"} /-->
+<div class="wp-block-group"><!-- wp:site-title {"level":0,"isLink":false,"style":{"typography":{"fontStyle":"normal","fontWeight":"600"}},"fontSize":"medium"} /-->
```
MaggieCabrera, line 12: I changed the font sizes to use preset values so the pattern works better in other themes too. To do this I updated theme.json with preset values that use the same slugs as core

poetry/patterns/header.php
```diff
@@ -9,16 +9,16 @@
...
 
 <!-- wp:group {"style":{"spacing":{"blockGap":"48px"}},"layout":{"type":"flex","orientation":"vertical"}} -->
-<div class="wp-block-group"><!-- wp:heading {"level":1} -->
-<h1>Food <br>for thoughts</h1>
```
MaggieCabrera, line 16: I removed headings from the header since they appear in almost all pages. We may want the h1 to be something else like the title of a page/post or the archive title instead. I used the preset font sizes instead to keep the same styling for the paragraphs

MaggieCabrera:
@sandroschutt I made a few changes to your PR and explained the reasoning behind them. I will bring this in, but we can iterate if we think it needs some adjustment!

MaggieCabrera (approved):

PR #142, opened 2024-03-05
Fix Poetry's missing single post template

### LINKED ISSUE #128: Poetry: Single posts template
The theme only designs an index template, we need a template for the single posts.

sandroschutt:
I've created the template and patter files for single posts.

### DIFF
--- a/poetry/patterns/single.php
+++ b/poetry/patterns/single.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * Title: single
+ * Slug: poetry/single
+ * Categories: hidden
+ * Inserter: no
+ */
+?>
+<!-- wp:template-part {"slug":"header","tagName":"header"} /-->
+
+<!-- wp:group {"layout":{"type":"constrained"}} -->
+<div class="wp-block-group"><!-- wp:group {"style":{"spacing":{"padding":{"top":"96px","right":"0","bottom":"96px","left":"0"},"blockGap":"40px"}},"layout":{"type":"flex","flexWrap":"nowrap","orientation":"horizontal","verticalAlignment":"top","justifyContent":"center"}} -->
+<div class="wp-block-group" style="padding-top:96px;padding-right:0;padding-bottom:96px;padding-left:0"><!-- wp:group {"style":{"spacing":{"padding":{"top":"96px"}}},"layout":{"type":"default"}} -->
+<div class="wp-block-group" style="padding-top:96px"><!-- wp:post-date /--></div>
+<!-- /wp:group -->
+
+<!-- wp:group {"style":{"spacing":{"padding":{"right":"56px","left":"56px"},"blockGap":"32px"},"border":{"left":{"width":"2px"}}},"layout":{"type":"flex","orientation":"vertical","flexWrap":"nowrap"}} -->
+<div class="wp-block-group" style="border-left-width:2px;padding-right:56px;padding-left:56px"><!-- wp:image {"id":15,"width":"56px","height":"56px","sizeSlug":"full","linkDestination":"none"} -->
+<figure class="wp-block-image size-full is-resized"><img src="<?php echo esc_url( get_stylesheet_directory_uri() ); ?>/assets/images/feather-small.png" alt="" class="wp-image-15" style="width:56px;height:56px"/></figure>
+<!-- /wp:image -->
+
+<!-- wp:post-title {"style":{"spacing":{"padding":{"right":"24%"}}}} /-->
+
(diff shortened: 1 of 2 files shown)


iamsam2e (commented):
Header and footer templates don't initialize on first render of the page, but removing & reincerting the component shows they works - looks to be due to working out of a subdirectory; cloning into /themes does make those components render on first page load in FSE editor

MaggieCabrera:
> Header and footer templates don't initialize on first render of the page, but removing & reincerting the component shows they works - looks to be due to working out of a subdirectory; cloning into /themes does make those components render on first page load in FSE editor

Themes in a subdirectory is a valid usecase, so we need to account for that. Good catch!

MaggieCabrera:
@samtoohey93 are you sure that you have the latest version installed? looking at the code, the template parts should be loading correctly

samtoohey93:
> @samtoohey93 are you sure that you have the latest version installed? looking at the code, the template parts should be loading correctly

Yep! running latest versions of everything, but i'll run another check :) 
Pls disregard in the meantime :smile:

iamsam2e:
Issue was template component was not rendering on first page load initially.

image

Upon investigation @sandroschutt it looks like removing the theme slug resolves the issue and allows the component to render correctly. 

``` 
<!-- wp:template-part {"slug":"header","tagName":"header"} /-->
``` 

Hope this helps!

thetwopct:
Related issue - https://github.com/WordPress/community-themes/issues/128

sandroschutt:
> Header and footer templates don't initialize on first render of the page, but removing & reincerting the component shows they works - looks to be due to working out of a subdirectory; cloning into /themes does make those components render on first page load in FSE editor

Not sure if we're in the same issue here. I tested it in my dev stage and just created a dedicated test stage for replicating the problem you mentioned, but it didn't show.

poetry/patterns/single.php
```diff
@@ -0,0 +1,51 @@
...
+ * Categories: hidden
+ * Inserter: no
+ */
+?>
+<!-- wp:template-part {"slug":"header","tagName":"header"} /-->
+
+<!-- wp:group {"layout":{"type":"constrained"}} -->
+<div class="wp-block-group"><!-- wp:group {"style":{"spacing":{"padding":{"top":"96px","right":"0","bottom":"96px","left":"0"},"blockGap":"40px"}},"layout":{"type":"flex","flexWrap":"nowrap","orientation":"horizontal","verticalAlignment":"top","justifyContent":"center"}} -->
+<div class="wp-block-group" style="padding-top:96px;padding-right:0;padding-bottom:96px;padding-left:0"><!-- wp:group {"style":{"spacing":{"padding":{"top":"96px"}}},"layout":{"type":"default"}} -->
+<div class="wp-block-group" style="padding-top:96px"><!-- wp:post-date /--></div>
```
MaggieCabrera, line 14: Let's use presets instead of hardcoded values for spacing

poetry/patterns/single.php
```diff
@@ -0,0 +1,51 @@
...
+<!-- /wp:group -->
+
+<!-- wp:group {"style":{"spacing":{"padding":{"right":"56px","left":"56px"},"blockGap":"32px"},"border":{"left":{"width":"2px"}}},"layout":{"type":"flex","orientation":"vertical","flexWrap":"nowrap"}} -->
+<div class="wp-block-group" style="border-left-width:2px;padding-right:56px;padding-left:56px"><!-- wp:image {"id":15,"width":"56px","height":"56px","sizeSlug":"full","linkDestination":"none"} -->
+<figure class="wp-block-image size-full is-resized"><img src="<?php echo esc_url( get_stylesheet_directory_uri() ); ?>/assets/images/feather-small.png" alt="" class="wp-image-15" style="width:56px;height:56px"/></figure>
```
MaggieCabrera, line 19: let's use `get_template_directory_uri` instead

poetry/patterns/single.php
```diff
@@ -0,0 +1,51 @@
...
+
+<!-- wp:group {"style":{"spacing":{"blockGap":"8px"},"elements":{"link":{"color":{"text":"var:preset|color|vivid-red"}}}},"layout":{"type":"flex","flexWrap":"nowrap"}} -->
+<div class="wp-block-group has-link-color"><!-- wp:paragraph {"textColor":"custom-contrast"} -->
+<p class="has-custom-contrast-color has-text-color">By</p>
+<!-- /wp:paragraph -->
+
+<!-- wp:post-author {"showAvatar":false,"showBio":false,"textColor":"custom-contrast"} /--></div>
+<!-- /wp:group -->
+
+<!-- wp:post-content /--></div>
```
MaggieCabrera, line 32: Should we add the comments block or is it ok not to have it for this theme?

MaggieCabrera (changes requested):

MaggieCabrera:
We have similar work done in https://github.com/WordPress/community-themes/pull/146 how can we conflate the two?

PR #143, opened 2024-03-06
Fix Poetry page template

sandroschutt:
Creates a template file for pages in Poetry.


MaggieCabrera:
Thanks for the PR! I think it´s better if we split the PRs to one per template for clarity of review

sandroschutt:
My bad. I think I messed up something while commiting in my terminal but didn't notice. Thanks for the heads up.

PR #145, opened 2024-03-07
Adding Poetry: 404 and search templates

### LINKED ISSUE #130: Poetry: 404 and search
The theme only designs an index template, we need a template for the search and 404

thetwopct:
https://github.com/WordPress/community-themes/issues/130

## 404

Screenshot-2024-03-07 --13 51 45

I gave it a poetry-esque 404 message.

## Search

Screenshot-2024-03-07 --14 07 17

The loop results uses the same style as the index page, but looks like this needs some adjustment to the column/group widths to stop the date wrapping?

### DIFF
--- a/poetry/templates/404.html
+++ b/poetry/templates/404.html
@@ -1,0 +1,29 @@
+<!-- wp:group {"style":{"spacing":{"padding":{"top":"32px","bottom":"32px"}}},"layout":{"type":"constrained"}} -->
+<div class="wp-block-group" style="padding-top:32px;padding-bottom:32px">
+    <!-- wp:group {"style":{"spacing":{"blockGap":"64px"}},"layout":{"type":"flex","flexWrap":"nowrap"}} -->
+    <div class="wp-block-group">
+        <!-- wp:group {"style":{"spacing":{"blockGap":"88px"}},"layout":{"type":"flex","orientation":"vertical"}} -->
+        <div class="wp-block-group">
+            <!-- wp:site-title {"level":0,"isLink":false,"style":{"typography":{"fontSize":"24px","fontStyle":"normal","fontWeight":"600"}},"textColor":"custom-text"} /-->
+
+            <!-- wp:group {"style":{"spacing":{"blockGap":"48px"}},"layout":{"type":"flex","orientation":"vertical"}} -->
+            <div class="wp-block-group"><!-- wp:heading {"level":1} -->
+                <h1 class="wp-block-heading">Alas, <br>where words once wandered</h1>
+                <!-- /wp:heading -->
+            </div>
+            <!-- /wp:group -->
+        </div>
+        <!-- /wp:group -->
+    </div>
+    <!-- /wp:group -->
+</div>
+<!-- /wp:group -->
+
+<!-- wp:group {"style":{"spacing":{"padding":{"top":"32px","bottom":"64px"}}},"layout":{"type":"constrained"}} -->
+<div class="wp-block-group" style="padding-top:32px;padding-bottom:64px"><!-- wp:heading {"level":5} -->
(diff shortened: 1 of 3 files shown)


poetry/templates/404.html
```diff
@@ -0,0 +1,29 @@
+<!-- wp:group {"style":{"spacing":{"padding":{"top":"32px","bottom":"32px"}}},"layout":{"type":"constrained"}} -->
+<div class="wp-block-group" style="padding-top:32px;padding-bottom:32px">
+    <!-- wp:group {"style":{"spacing":{"blockGap":"64px"}},"layout":{"type":"flex","flexWrap":"nowrap"}} -->
+    <div class="wp-block-group">
+        <!-- wp:group {"style":{"spacing":{"blockGap":"88px"}},"layout":{"type":"flex","orientation":"vertical"}} -->
+        <div class="wp-block-group">
+            <!-- wp:site-title {"level":0,"isLink":false,"style":{"typography":{"fontSize":"24px","fontStyle":"normal","fontWeight":"600"}},"textColor":"custom-text"} /-->
```
MaggieCabrera: I think it would be better to keep the header element like the index.html template has it. It's more consistent between pages and also we need to keep the `header` element for accessibility reasons. This comment goes for both templates

poetry/templates/search.html
```diff
@@ -0,0 +1,77 @@
...
+            <!-- /wp:group -->
+            <!-- /wp:post-template -->
+        </div>
+        <!-- /wp:group -->
+    </main>
+    <!-- /wp:query -->
+</main>
+<!-- /wp:group -->
+
+<!-- wp:template-part {"slug":"footer","theme":"poetry","tagName":"footer","area":"footer"} /-->
```
MaggieCabrera, line 77: Both templates should have a footer element. We can go without the area attribute (the theme.json file already includes this info) and the theme slug (it will make the template part break if the theme is installed in a subfolder)

poetry/templates/search.html
```diff
@@ -0,0 +1,77 @@
...
+</div>
+<!-- /wp:group -->
+
+<!-- wp:spacer {"height":"50px"} -->
+<div style="height:50px" aria-hidden="true" class="wp-block-spacer"></div>
+<!-- /wp:spacer -->
+
+<!-- wp:group {"tagName":"main","backgroundColor":"custom-background-light","layout":{"contentSize":null,"type":"constrained"}} -->
+<main class="wp-block-group has-custom-background-light-background-color has-background">
+    <!-- wp:query {"queryId":0,"query":{"perPage":10,"pages":0,"offset":0,"postType":"post","order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":true,"taxQuery":null,"parents":[]},"tagName":"main","layout":{"contentSize":null,"type":"constrained"}} -->
```
MaggieCabrera, line 27: Since this Query block should look the same as the one in the index template, what about we make it into a pattern instead and then we avoid duplicating code over both templates? (that way when we fix the problem with the date we only change it in one place)

MaggieCabrera (changes requested):
This is looking nice! I added a few comments. I think we can follow up on the issue with the date on a separate PR.

poetry/templates/404.html
MaggieCabrera: Ok, I forgot the header in poetry was super huge. My recommendation still remains, let's make a secondary header pattern or template part to reuse in internal pages that doesn't have the huge text, but let's keep it consistent between pages. And let's use the tagName header on it so we don't miss the header element

MaggieCabrera:
@samtoohey93 be careful here because this needs a rebase

iamsam2e:
Ah damn sorry I was doing this rushing out the door this arvo 🤦‍♂️
I'm back in office Monday, I'll fix this. Apologies.

PR #146, opened 2024-03-07
singular page template scaffold

### LINKED ISSUE #128: Poetry: Single posts template

iamsam2e:
version 1 basic scaffold (tested in editor)

### DIFF
--- a/poetry/templates/singular.html
+++ b/poetry/templates/singular.html
@@ -1,4 +1,4 @@
-<!-- wp:template-part {"slug":"header","theme":"community-themes/poetry"} /-->
+<!-- wp:template-part {"slug":"header","tagName":"header"} /-->
 <!-- wp:group {"tagName":"main","align":"full"} -->
 <main class="wp-block-group alignfull"><!-- wp:group {"style":{"spacing":{"padding":{"top":"var:preset|spacing|50"},"margin":{"bottom":"var:preset|spacing|40"}}},"layout":{"type":"constrained"}} -->
     <div class="wp-block-group" style="margin-bottom:var(--wp--preset--spacing--40);padding-top:var(--wp--preset--spacing--50)"><!-- wp:post-featured-image {"style":{"spacing":{"margin":{"bottom":"var:preset|spacing|40"}}}} /-->
--- a/poetry/templates/singular.html
+++ b/poetry/templates/singular.html
@@ -1,0 +1,31 @@
+<!-- wp:template-part {"slug":"header","theme":"community-themes/poetry"} /-->
+<!-- wp:group {"tagName":"main","align":"full"} -->
+<main class="wp-block-group alignfull"><!-- wp:group {"style":{"spacing":{"padding":{"top":"var:preset|spacing|50"},"margin":{"bottom":"var:preset|spacing|40"}}},"layout":{"type":"constrained"}} -->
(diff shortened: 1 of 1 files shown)


iamsam2e:
Editors note: apologies its been a white since PR'ing to a project this big, im too used to smaller project use of Git.

In reference to work on #142 from Theme session @ WCAsia
Scaffold adds header, empty featured image and title block to start the basic SPT scaffold

image

poetry/templates/singular.html
```diff
@@ -0,0 +1,31 @@
+<!-- wp:template-part {"slug":"header","theme":"community-themes/poetry"} /-->
```
MaggieCabrera:
```suggestion
<!-- wp:template-part {"slug":"header","tagName":"header"} /-->
```
We need the header tagName for accessibility, and we need to remove the theme slug because that route may not be the same for every installation

poetry/templates/singular.html
```diff
@@ -0,0 +1,31 @@
...
+    <hr class="wp-block-separator has-text-color has-contrast-3-color has-alpha-channel-opacity has-contrast-3-background-color has-background is-style-wide" style="margin-bottom:var(--wp--preset--spacing--40)"/>
+    <!-- /wp:separator -->
+    
+    
+    <!-- wp:pattern {"slug":"poetry/hidden-post-navigation"} /--></div>
+    <!-- /wp:group --></div>
+    <!-- /wp:group --></main>
+    <!-- /wp:group -->
+    
+    <!-- wp:template-part {"slug":"footer","tagName":"footer","area":"footer"} /-->
```
MaggieCabrera, line 31:
```suggestion
    <!-- wp:template-part {"slug":"footer","tagName":"footer"} /-->
```
We don't need to define the area because it's already defined in theme.json (or should be!)

poetry/templates/singular.html
```diff
@@ -0,0 +1,31 @@
...
+    <div class="wp-block-group"><!-- wp:spacer {"height":"var:preset|spacing|40"} -->
+    <div style="height:var(--wp--preset--spacing--40)" aria-hidden="true" class="wp-block-spacer"></div>
+    <!-- /wp:spacer -->
+    
+    <!-- wp:separator {"style":{"spacing":{"margin":{"bottom":"var:preset|spacing|40"}}},"backgroundColor":"contrast-3","className":"is-style-wide"} -->
+    <hr class="wp-block-separator has-text-color has-contrast-3-color has-alpha-channel-opacity has-contrast-3-background-color has-background is-style-wide" style="margin-bottom:var(--wp--preset--spacing--40)"/>
+    <!-- /wp:separator -->
+    
+    
+    <!-- wp:pattern {"slug":"poetry/hidden-post-navigation"} /--></div>
```
MaggieCabrera, line 26: This doesn't exist, we should create it! It's fine if we add it to this PR

MaggieCabrera (changes requested):
This is looking really solid. I added a few comments.

One more thing to keep in mind: by deciding to make this a singular template instead of single, we are using the same template for pages and posts, which means posts will not have a comments block. It could be a design decision for this theme, but it feels a little heavy-handed to me.

MaggieCabrera:
We have similar work done in https://github.com/WordPress/community-themes/pull/142 how can we conflate the two?

PR #150, opened 2024-03-18
Fixes the two first tasks on issue #126

### LINKED ISSUE #126: Purr: overview tasks
There are a few things that this theme needs to improve before launch. Feel free to open a PR for each of the items or open further issues if you find other problems:

- [ ] Templates have hardcoded colors, they should live in theme.json instead
- [ ] Same happens with spacing, let's figure out some global spacing values that we can use for the templates instead of hardcoded values
- [ ] The style.css needs changes. The author needs to be the wordpress.org user (let´s copy from blue note or  stacks). The contributors that work on the theme will get a badge but the user right now can only be one, so all our themes have the same author. The WordPress versions need updating too
- [ ] After the design changes, we need to update the screenshot

sandroschutt:
Related issue - #126 

I removed most of the hard code in templates, patterns and parts. Couldn't manage to remove the sidebars position attributes, though.


MaggieCabrera:
I don't see the changes clearly in this diff, I think ti's best if we leave formatting to a separate PR so it's easier to review the specific change that this PR is trying to solve

MaggieCabrera:
@sandroschutt can you update your PRs to only include the changes that we are meant to review, please?

PR #152, opened 2024-03-25
Poetry: fix index template

### LINKED ISSUE #138: Poetry: header not working
The header of the theme is not showing up

MaggieCabrera:
This PR makes a few fixes for the index template:

- Turns the pattern into only a query block pattern, making it visible in the inserter and reusable for other query blocks. We can then reuse it for the 404 and search pages.
- Fixes the header and footer so they work regardless of where the theme is installed
- Fixed the layout so the date container is always the same size

Closes #138


poetry/patterns/posts.php
```diff
@@ -1,24 +1,25 @@
...
 <!-- /wp:group -->
 
+<!-- wp:group {"style":{"layout":{"selfStretch":"fill","flexSize":null}},"layout":{"type":"constrained"}} -->
+<div class="wp-block-group">
 <!-- wp:group {"style":{"spacing":{"padding":{"right":"56px","left":"56px"},"blockGap":"32px"},"border":{"left":{"width":"2px"}}},"layout":{"type":"flex","orientation":"vertical","flexWrap":"nowrap"}} -->
-<div class="wp-block-group" style="border-left-width:2px;padding-right:56px;padding-left:56px"><!-- wp:image {"id":15,"width":56,"height":56,"sizeSlug":"full","linkDestination":"none"} -->
-<figure class="wp-block-image size-full is-resized"><img src="<?php echo esc_url( get_stylesheet_directory_uri() ); ?>/assets/images/feather-small.png" alt="" class="wp-image-15" width="56" height="56"/></figure>
+<div class="wp-block-group" style="border-left-width:2px;padding-right:56px;padding-left:56px"><!-- wp:image {"width":56,"height":56,"sizeSlug":"full","linkDestination":"none"} -->
+<figure class="wp-block-image size-full is-resized"><img src="<?php echo esc_url( get_stylesheet_directory_uri() ); ?>/assets/images/feather-small.png" alt="" width="56" height="56"/></figure>
```
shail-mehta: Can We use `get_template_directory_uri()` instead of `get_stylesheet_directory_uri()` ?  

Because Poetry theme is main theme.

shail-mehta (changes requested):

poetry/patterns/posts.php
MaggieCabrera: thank you for pointing that one out!

PR #155, opened 2024-03-28
Used get_template_directory_uri() instead of get_stylesheet_directory_uri()

shail-mehta:
Poetry & Blue Note both are main theme 

-  We need to add get_template_directory_uri() instead of get_stylesheet_directory_uri(). 
- Also removed Image ids & classes in images

**For Ex:** 
- "id":15
- class="wp-image-15"


blue-note/patterns/media-text.php
```diff
@@ -6,8 +6,8 @@
  * Block Types: core/media-text
  */
 ?>
-<!-- wp:media-text {"align":"wide","mediaId":146,"mediaLink":"<?php echo esc_url( get_stylesheet_directory_uri() ); ?>/assets/images/John_Coltrane.png","mediaType":"image","mediaWidth":41,"verticalAlignment":"top","imageFill":true,"style":{"spacing":{"padding":{"right":"0","left":"0"}}}} -->
```
MaggieCabrera, line 9: This pattern is showing up broken on the site editor, but the code changes make sense to me, there must be something else

MaggieCabrera (changes requested):
Can you remove the changes to Poetry? They are already included in https://github.com/WordPress/community-themes/pull/152.

If you do that and then change the version number and update the changelog for Blue Note I'll upload the theme with the changes after the merge. Thank you for this!

blue-note/readme.txt
```diff
@@ -11,6 +11,9 @@ Blue Note is an elegant and fun block theme inspired by jazz and the record labe
 
 == Changelog ==
 
+= 1.0.2 =
+* Replaced get_stylesheet_directory_uri() with get_template_directory_uri()
```
MaggieCabrera:
```suggestion
* Replaced get_stylesheet_directory_uri() with get_template_directory_uri() to ensure child themes work correctly
```

blue-note/patterns/media-text.php
MaggieCabrera: I tested and the problem is with removing the id of the media-text block. Maybe we should open a Gutenberg issue to see why the block markup fails without them. Let's return them so the pattern goes back to working

shail-mehta: Yeah I just checked with id  and it works fine.

I will revert this pattern's id and class related changes.

shail-mehta: I have added Image id & class in this pattern & now its working fine for me.

MaggieCabrera: Fantastic, great job!

MaggieCabrera (approved):

PR #159, opened 2024-04-05
Added template 404 in Purr theme

pablo-miralles:
This commit adds a 404 template in Purr theme (community-themes/templates/404.html)


MaggieCabrera:
This is looking good. There's a lot of repeated code but we can fix that in a follow up

MaggieCabrera (approved):

PR #162, opened 2024-04-15
Add links to previews for theme changes using playground

MaggieCabrera:
This PR blatantly copies the implementation in the Automattic themes repo (https://github.com/Automattic/themes/pull/7719). Thank you @vcanales for this, it will work nicely for community themes as well.

This introduces a Github action that runs on pull requests and adds a link to preview the changes using Playground. I made changes to Atlas and Poetry, and the links can be seen on this same PR.

### DIFF
--- a/.github/workflows/preview-theme.yml
+++ b/.github/workflows/preview-theme.yml
@@ -39,7 +39,7 @@ jobs:
           # Output list of theme slugs with changes
           echo "HAS_THEME_CHANGES=true" >> $GITHUB_OUTPUT
           echo "CHANGED_THEMES=$(echo ${unique_dirs[@]})" >> $GITHUB_ENV
-          echo "Theme directories with changes: $CHANGED_THEME_DIRS"
+          echo "Theme directories with changes: $CHANGED_THEMES"
       - name: Add Preview Links comment
         id: comment-on-pr
         if: ${{ steps.check-for-changes.outputs.HAS_THEME_CHANGES == 'true' }}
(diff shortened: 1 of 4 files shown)


.github/workflows/preview-theme.yml
```diff
@@ -0,0 +1,73 @@
...
+          # Check if themes have changed
+          if [[ ${#unique_dirs[@]} -eq 0 ]]; then
+              echo "No theme changes detected"
+              echo "HAS_THEME_CHANGES=false" >> $GITHUB_OUTPUT
+              exit 78 # Use neutral exit code
+          fi
+          # Output list of theme slugs with changes
+          echo "HAS_THEME_CHANGES=true" >> $GITHUB_OUTPUT
+          echo "CHANGED_THEMES=$(echo ${unique_dirs[@]})" >> $GITHUB_ENV
+          echo "Theme directories with changes: $CHANGED_THEME_DIRS"
```
vcanales:
```suggestion
          echo "Theme directories with changes: $CHANGED_THEMES"
```

There's a mismatch between the variable names here.
I added that line as debug, but I kept it in because it's nice to have that kind of info in the CI; it's not functionally necessary though, in case you'd like to remove it.

vcanales (approved):
LGTM!

.github/workflows/preview-theme.yml
MaggieCabrera: good catch! and thank you for the script :D

MaggieCabrera:
@vcanales when I removed the theme changes it failed, hehe

looks like `exit 78` makes the fail check, when it should pass

Issue #164, opened 2024-04-18
Archivist: a theme of default styles


karmatosed:
I have a new theme to work on and would appreciate the opportunity of collaborating. It is a simple concept:

- A theme.
- Styles that pay homage to default themes. For example one for TwentyTen, TwentyEleven... all the way up to Twenty Twenty-Four.
- A collection of patterns and templates that also pay homage to default themes.

As I share this I want to draw attention to the awesome suggestion by @carolinan to create ‘[Block theme copies of the classic default themes’](https://github.com/WordPress/community-themes/issues/163). If this is decided to also create as an approach it would be great. I think the more options the better. 

I would also be incredibly grateful to have collaboration on Tardis, in that respect I will be uploading an initial theme start as a PR to attach to this. It won’t be anything more than a foundation and we can then work collaboratively.

I hope others will join me in this and want to also support any other default theme explorations going on - it’s great there are others. 

Thank you @MaggieCabrera for the suggestion of adding this to the repo as I truly feel it is the right spot for it and excited to dive in.

## Homage not exact
This phrase is key. The styling needs to be just that, a style not a direct copy or design from the original. This is going to lead to some fun exploring creatively where we can dive into things. What is the essence of Twenty Ten or Twenty Thirteen for example we can interpret today into a style?

## Next steps
I will be adding the initial commit as a PR and then create issues to start work on each aspect.

## Feedback requests
I am currently looking for feedback on the approach, ideas for which themes to work on, what patterns could be created or templates and also people who are interested in being involved. As far as involvement goes, I plan to create issues for each piece that is worked on so there will be opportunity for anything. If there is something specifically you want to take on please say.

PR #165, opened 2024-04-18
Initial push of foundation for Tardis with no styling

karmatosed:
This is an initial commit using Create Block theme as a foundation to start the collaboration for Tardis. What this has is:

- Foundation theme.
- Style folder

There will be issues created to work on specific aspects of this, it is just to start the development as per issue #164


tardis/styles/twentytwentyfour.json
```diff
@@ -0,0 +1,7 @@
+{
+	"$schema": "https://schemas.wp.org/wp/6.5/theme.json",
+	"version": 2,
+	"title": "TwentyTwentyFour",
```
MaggieCabrera, line 4: Not a blocker for this but the title might look bad on the UI if we write them like this

MaggieCabrera (approved):
Approving so we can get the ball rolling with this one

carolinan:
My only concern is the trademark registered on the name TARDIS. Would it cover this?

MaggieCabrera:
> My only concern is the trademark registered on the name TARDIS. Would it cover this?

We should investigate that, but it's not a blocker, so I'm merging this. Let's revisit that when the theme is ready for publishing

PR #170, opened 2024-04-27
Revert changes in preview-themes workflow

vcanales:
Reverts back to using `pull_request` as the event trigger for the preview themes workflow.


mikachan (approved):
These changes look good to me - I can see the event trigger has been changed to `pull_request` 👍

PR #172, opened 2024-05-05
Archivist: Updates foundation colors and also basics

karmatosed:
This pull brings in a foundation color system using Twenty Twenty-Four to start and also adds some basic values to get the theme started.

### DIFF
--- a/archivist/theme.json
+++ b/archivist/theme.json
@@ -3,9 +3,68 @@
 	"version": 2,
 	"settings": {
 		"appearanceTools": true,
+		"color": {
+			"defaultDuotone": false,
+			"defaultGradients": false,
+			"defaultPalette": false,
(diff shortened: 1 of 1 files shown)


archivist/theme.json
```diff
@@ -3,9 +3,68 @@
 	"version": 2,
 	"settings": {
 		"appearanceTools": true,
+		"color": {
+			"defaultDuotone": false,
+			"defaultGradients": false,
+			"defaultPalette": false,
+			"duotone": [],
```
MaggieCabrera, line 10: Should we add some?

MaggieCabrera (approved):
I'm fine bringing this in, but let's add a variation specific to TT4 and update the default theme.json with new colors?

PR #179, opened 2024-05-17
Poetry: add alternative header part

### LINKED ISSUE #174: Poetry: Create alternative header
Let's create a smaller header for pages like 404 and search

adrianaleites:
- Added a smaller alternative header for pages like 404 and search.
- Added functions.php to enqueue styles.

Closes #174


poetry/parts/header-small.html
```diff
@@ -0,0 +1,25 @@
+<!-- wp:group {"style":{"spacing":{"padding":{"top":"32px","bottom":"64px"}}},"className":"header-small","layout":{"type":"constrained"}} -->
```
MaggieCabrera: Instead of pixel values, can we use any of the presets (small, medium, large...)? If none work, let's use rem values instead

poetry/parts/header-small.html
```diff
@@ -0,0 +1,25 @@
...
+<p class="has-rozha-one-font-family" style="font-size:2rem;line-height:0.8">Food for thoughts</p>
+<!-- /wp:paragraph -->
+
+<!-- wp:paragraph {"style":{"typography":{"lineHeight":"1.5"}},"fontSize":"small"} -->
+<p class="has-small-font-size" style="line-height:1.5">A collection of valuable thoughts by a very egocentric person.</p>
+<!-- /wp:paragraph --></div>
+<!-- /wp:group -->
+
+<!-- wp:image {"width":"auto","height":"64px","sizeSlug":"large","linkDestination":"none"} -->
+<figure class="wp-block-image size-large is-resized"><img src="http://wp-community-day.local/wp-content/themes/poetry/assets/images/feather.png" alt="" style="width:auto;height:64px"/></figure>
```
MaggieCabrera: this url won't work on other people's sites, let's do something like [this](https://github.com/WordPress/community-themes/blob/dfce47dce46317c9521a95fd86dc117924696041/poetry/patterns/header.php#L26)

poetry/parts/header-small.html
MaggieCabrera: I just checked and the old header has the same problem, we should do the change on both to keep it consistent

adrianaleites:
@MaggieCabrera Thanks for your review. I will try to address your comments here regarding the use of pixel values. I will add the spacingSizes in theme.json as in twentytwentyfour. And I will also convert the header-small template to be a pattern instead.

But I just notice that the page navigation is missing so maybe I will have to come up with a different solution for the small version of the header that can accomodate the navigation. There's no navigation in the original header pattern also. Or maybe It's supposed to be like that and this specific theme don't need a navigation in the header. What do you think?

MaggieCabrera:
> @MaggieCabrera Thanks for your review. I will try to address your comments here regarding the use of pixel values. I will add the spacingSizes in theme.json as in twentytwentyfour. And I will also convert the header-small template to be a pattern instead.
> 
> But I just notice that the page navigation is missing so maybe I will have to come up with a different solution for the small version of the header that can accomodate the navigation. There's no navigation in the original header pattern also. Or maybe It's supposed to be like that and this specific theme don't need a navigation in the header. What do you think?

I think we can make of it what we like. It makes sense to me that the focus of this kind of blog would be the post content, but if you click on internal pages you would like to go back to the home. Maybe we want our smaller header to just have the logo and the page title, removing the text "A collection of valuable thoughts...". In fact, now that I look at it, we should probably replace the image block for the site logo block and the other paragraph that we have for the "tagline" block instead.

adrianaleites:
@MaggieCabrera I'm so sorry to have abandoned this PR. Right after I had open it I got sick. And then, got to focus on work for some time. And then, came holidays and I totally forgot to get back here. I see you are trying to merge this into another PR. Tell me if I can help in some way. I will also check the current open issues and see if I can help with any.

MaggieCabrera:
> @MaggieCabrera I'm so sorry to have abandoned this PR. Right after I had open it I got sick. And then, got to focus on work for some time. And then, came holidays and I totally forgot to get back here. I see you are trying to merge this into another PR. Tell me if I can help in some way. I will also check the current open issues and see if I can help with any.

Hi! so happy to see you back. I've been busy myself, so don't worry. I think both PRs are a little stale, so if you want to pick up the work it should be fine. Maybe from a fresh PR and we can close the other 2?

MaggieCabrera:
Closing in favor of #235

Issue #104, opened 2023-08-11
Term: Single post


MaggieCabrera:
![Post](https://github.com/WordPress/community-themes/assets/3593343/ed5a6a0c-63ce-408a-9b61-4d27256404bc)
![Post with Featured Image](https://github.com/WordPress/community-themes/assets/3593343/7ce26ac7-be88-42e3-833d-25b89879234f)
https://www.figma.com/community/file/1263420783737936752
Let's repeat the same design without the comments/meta block for the page template

PR #182, opened 2024-05-17
Single post template to Term theme

mlaetitia:
Issue #104
Figma: https://www.figma.com/community/file/1263420783737936752/console

Adds the single post template to the Term theme.


MaggieCabrera (approved):
this is a good starting point, thank you!

PR #184, opened 2024-05-17
Term: Updates the global styles: color palette, typography

mlaetitia:
Changes the color palette to include the colors in Figma and typography:
- Heading font sizes
- Font sizes for small, regular, ...
- Default capitalization
- Default font

### DIFF
--- a/term/theme.json
+++ b/term/theme.json
@@ -83,34 +83,34 @@
 		"elements": {
 			"h1": {
 				"typography": {
-					"fontSize": "64px",
+					"fontSize": "4rem",
 					"fontStyle": "normal",
 					"fontWeight": "700"
 				}
 			},
 			"h2": {
 				"typography": {
-					"fontSize": "48px"
+					"fontSize": "3rem"
 				}
 			},
 			"h3": {
 				"typography": {
-					"fontSize": "38px"
+					"fontSize": "2.38rem"
 				}
 			},
 			"h4": {
 				"typography": {
-					"fontSize": "28px"
+					"fontSize": "1.75rem"
 				}
 			},
 			"h5": {
 				"typography": {
-					"fontSize": "22px"
+					"fontSize": "1.38rem"
 				}
 			},
 			"h6": {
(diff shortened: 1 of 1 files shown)


term/theme.json
```diff
@@ -18,16 +37,95 @@
...
+			"h1": {
+				"typography": {
+					"fontSize": "64px",
+					"fontStyle": "normal",
+					"fontWeight": "700"
+				}
+			},
+			"h2": {
+				"typography": {
+					"fontSize": "48px"
```
MaggieCabrera: can any of these use preset values? if not, let's use rem values instead so they can scale

term/theme.json
```diff
@@ -18,16 +37,95 @@
...
+				}
+			},
+			"link": {
+				"color": {
+					"text": "var(--wp--preset--color--custom-grey)"
+				}
+			}
+		},
 		"spacing": {
+			"blockGap": "25px",
```
MaggieCabrera: rem units are better for spacing too

term/theme.json
```diff
@@ -36,7 +134,10 @@
 			}
 		},
 		"typography": {
-			"fontFamily": "var(--wp--preset--font-family--system-font)"
+			"fontFamily": "var(--wp--preset--font-family--system-font)",
+			"fontSize": "var(--wp--preset--font-size--medium)",
+			"lineHeight": "1.5",
+			"textTransform": "capitalize"
```
MaggieCabrera, line 140: I think the body font weight needs to be thinner (but maybe it's my mac, that doesn't have helvetica!)

mlaetitia: On mine, I can see the font in a size size, but I can definitely change it back as it was, or not set a default. What do you think is the best way here ?

mlaetitia:
@MaggieCabrera Thanks for the review! Addressed your comments and happy to make any needed changes!

MaggieCabrera (approved):
Let's bring this in and tweak it as we go

PR #187, opened 2024-05-21
fix workflow forks

MaggieCabrera:
This tries to fix the workflow when the PR is submitted by a user without enough permissions via a fork


MaggieCabrera:
I'm merging this one to be able to test it with a fork

PR #144, opened 2024-03-07
Created a 404 error page

Soniabajpai:
Adding 404 page

PR #189, opened 2024-05-30
Update Fix

iamsam2e:
Rebased from PR 145, to resolve issues for the 404 template as suggested my MC as well as resolving accidental push onto the existing #145 branch

this resolves both #144 and #145


poetry/templates/404.html
```diff
@@ -1,3 +1,34 @@
+<<<<<<< HEAD
```
MaggieCabrera: this is a conflict :D

iamsam2e:
Going to try to address this this week! :)

MaggieCabrera:
I like the design and the text, but why are we making another header? do we think it's going to be one we want to reuse for other templates? Also, it might be time to move the parts that have text into a pattern for internationalization (there's more info on the readme file and you can use [blue note as a reference](https://github.com/WordPress/community-themes/blob/trunk/blue-note/templates/404.html))

PR #205, opened 2024-06-13
fixed typography in theme.json

maurodf0:
Removed capitlized and lowercase to heading and paragraph in theme.json


MaggieCabrera (approved):
This looks good. For future reference: these were added because there is capitalization on some of the templates in the design but that doesn't warrant making headings always lowercase and paragraphs always lowercase. Let's do that on a case by case basis on the specific templates or patterns

MaggieCabrera (approved):

PR #206, opened 2024-06-13
Term: Added rounded icon style for social links

tdrayson:
Registers social icon style for term


MaggieCabrera (approved):
Beautiful, thank you!

PR #208, opened 2024-06-13
Pattern: social in content

### LINKED ISSUE #108: Term: Pattern Social in content
Block Pattern_ Social In Content
https://www.figma.com/community/file/1263420783737936752

tdrayson:
Creats social in content pattern. (Requires rounded icons block style #206)
Screenshot 2024-06-13 at 12 26 30

### DIFF
--- a/term/patterns/social-in-content.php
+++ b/term/patterns/social-in-content.php
@@ -1,0 +1,29 @@
+<?php
+/**
+ * Title: Social in content
+ * Slug: term/social-in-content
+ * Categories: text
+ * Block Types: core/social-links
+ */
+<div class="wp-block-group" style="padding-top:0;padding-bottom:0"><!-- wp:columns {"style":{"spacing":{"blockGap":{"top":"0","left":"var:preset|spacing|60"},"margin":{"top":"0","bottom":"0"}}}} -->
+<div class="wp-block-columns" style="margin-top:0;margin-bottom:0"><!-- wp:column {"width":"50px"} -->
+<div class="wp-block-column" style="flex-basis:50px"><!-- wp:social-links {"iconColor":"custom-beige","iconColorValue":"#edece8","iconBackgroundColorValue":"#000","style":{"spacing":{"padding":{"top":"0","bottom":"0"},"margin":{"top":"0","bottom":"0"},"blockGap":{"top":"var:preset|spacing|40"}}},"className":"is-style-social-rounded-icons","layout":{"type":"flex","orientation":"vertical","justifyContent":"center","flexWrap":"nowrap"}} -->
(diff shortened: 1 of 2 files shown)


term/patterns/social-in-content.php
```diff
@@ -0,0 +1,29 @@
+<?php
+/**
+ * Title: Social in content
+ * Slug: term/social-in-content
+ * Categories: text
+ * Block Types: core/social-links
+ */
+<div class="wp-block-group" style="padding-top:0;padding-bottom:0"><!-- wp:columns {"style":{"spacing":{"blockGap":{"top":"0","left":"var:preset|spacing|60"},"margin":{"top":"0","bottom":"0"}}}} -->
+<div class="wp-block-columns" style="margin-top:0;margin-bottom:0"><!-- wp:column {"width":"50px"} -->
+<div class="wp-block-column" style="flex-basis:50px"><!-- wp:social-links {"iconColor":"custom-beige","iconColorValue":"#edece8","iconBackgroundColorValue":"#000","style":{"spacing":{"padding":{"top":"0","bottom":"0"},"margin":{"top":"0","bottom":"0"},"blockGap":{"top":"var:preset|spacing|40"}}},"className":"is-style-social-rounded-icons","layout":{"type":"flex","orientation":"vertical","justifyContent":"center","flexWrap":"nowrap"}} -->
```
MaggieCabrera: let's use the preset colors of the theme instead of the hex values. This makes it so if we add style variations, the colors will update when the presets are changed by the variation.

If the theme doesn't have a color palette in theme.json, then that means we should add one!

MaggieCabrera (changes requested):
A few details and this should be good to go! Add your name to the credits file if it's not there yet please

term/patterns/social-in-content.php
```diff
@@ -0,0 +1,29 @@
+<?php
+/**
+ * Title: Social in content
+ * Slug: term/social-in-content
+ * Categories: text
+ * Block Types: core/social-links
```
MaggieCabrera: we only add this when the pattern is styling just one block, mainly blocks that have variations like the query block, so we can remove this line in this case

term/patterns/social-in-content.php
```diff
@@ -0,0 +1,29 @@
+<?php
+/**
+ * Title: Social in content
+ * Slug: term/social-in-content
+ * Categories: text
```
MaggieCabrera: I would add call to action too

tdrayson: Added in latest commit

term/patterns/social-in-content.php
tdrayson: Removed in latest commit

term/patterns/social-in-content.php
tdrayson: Fixed in latest commit

Issue #199, opened 2024-06-05
Poetry: Variations


MaggieCabrera:
Let's add some variations to the theme

ironicmoka:
I'll try to work on this one today

PR #210, opened 2024-06-13
New Poetry variations and header font fixing (#199)

ironicmoka:
New variations for Poetry theme and a small fix to the header font handling


MaggieCabrera:
Can we get a different name for the "My typo" variation? These look nice. The variations are highlighting some issues that the theme has with padding and the little feather icon, but we can handle those separately

poetry/patterns/header.php
```diff
@@ -12,8 +12,8 @@
 <div class="wp-block-group"><!-- wp:site-title {"level":0,"isLink":true,"style":{"typography":{"fontStyle":"normal","fontWeight":"600"}},"fontSize":"medium"} /-->
 
 <!-- wp:group {"style":{"spacing":{"blockGap":"48px"}},"layout":{"type":"flex","orientation":"vertical"}} -->
-<div class="wp-block-group"><!-- wp:paragraph {"style":{"typography":{"lineHeight":"0.8"}},"fontSize":"xx-large","fontFamily":"rozha-one"} -->
-<p class="has-rozha-one-font-family has-xx-large-font-size" style="line-height:0.8">Food <br>for thoughts</p>
+<div class="wp-block-group"><!-- wp:paragraph {"style":{"typography":{"lineHeight":"0.8"}},"fontSize":"xx-large","fontFamily":"heading"} -->
+<h2 class="has-heading-font-family has-xx-large-font-size" style="line-height:0.8">Food <br>for thoughts</h2>
```
MaggieCabrera, line 16: we made this change but then we don't assign a "heading" font family in the .json files. The class `has-heading-font-family` is not resolving to anything so this change is not really doing anything. We could remove the class altogether and just let the theme default to whatever the theme.json sets for all headings. That is probably the best case for this particular theme, but that means that no templates/patterns can select specific font families and expect the theme to be able to adapt when we change to a different variation that doesn't have that specific font family.

shail-mehta:
@ironicmoka, could you please add the font license? Also, could you please update as suggested changes as above?

PR #211, opened 2024-06-13
Add Black Bordered Radius Block Variation and Pattern About Page

maurodf0:
Add register block style for image with black bordered radius image.


term/functions.php
```diff
@@ -34,6 +34,18 @@ function term_block_styles() {
 				}',
 			)
 		);
+
+		register_block_style(
+            'core/image',
+            array(
+                'name' => 'bordered',
+                'label' => __('Bordered', 'term'),
+                'inline_style' => '.wp-block-image.is-style-bordered img{
```
MaggieCabrera, line 43: I don't think we should do this because we can already use the border controls on the image block. We only did it for the social icons because those controls are not there yet

MaggieCabrera: The reason being: a user can only opt in or out of the border radius, they can't tweak it to other values

maurodf0: Yeah, you're right, don't think about it, sorry.

PR #216, opened 2024-06-20
Update readme files and theme json files to support proper WP versions

MaggieCabrera:
This PR updates all the theme.json and readme files to use the proper WordPress version. We shouldn't be using trunk on the schema for themes that are released (or about to be) because the trunk one will have features that are not on the minimum required version of WordPress required by the theme.

### DIFF
--- a/archivist/readme.txt
+++ b/archivist/readme.txt
@@ -1,8 +1,8 @@
 === Archivist ===
-Contributors: the WordPress team
+Contributors: wordpressdotorg
 Requires at least: 6.5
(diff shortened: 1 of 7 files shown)


MaggieCabrera:
I left TT1 blocks out of this since it doesn't have a schema and we should discuss what would be the required version for it when we finally get to update it

iamsam2e:
Seems all good - going to approve merge 👍

iamsam2e (approved):

PR #219, opened 2024-06-23
Created Pattern for Footer, Right Sidebar in Purr Theme

### LINKED ISSUE #192: Purr: Create patterns to avoid repetition between templates
The columns could be patterns and/or template parts to avoid repetition between templates.

If we think the user shouldn't be able to reuse these sections, the solution is to use a hidden pattern. If we decide it's a good idea for the user to reuse them we need to decide between pattern or template parts. If we want the user to be able to edit these sections and have those changes propagate to everywhere those sections are used, we would use template parts (because synced patterns are something that can't be done on a theme level yet, only the user level). If we think every instance of these sections is independent, then we would go for a normal pattern.

shail-mehta:
I have Created Pattern for Footer, Right Sidebar in Purr Theme.

Fixes: https://github.com/WordPress/community-themes/issues/192


purr/patterns/footer.php
```diff
@@ -0,0 +1,31 @@
+<?php
+/**
+ * Title: Footer
+ * Slug: purr/footer
+ * Categories: footer
+ * Description: A Footer section.
+ */
```
MaggieCabrera, line 8: we could add  `Block Types: core/template-part/footer` here

purr/patterns/right-sidebar.php
```diff
@@ -0,0 +1,15 @@
+<?php
+/**
+ * Title: Right Sidebar
+ * Slug: purr/right-sidebar
+ * Categories: sidebar
+ */
```
MaggieCabrera, line 9: I wonder if  `Block Types: core/template-part/sidebar` would work? I don't know if we have a replace flow for it or not

purr/patterns/right-sidebar.php
```diff
@@ -3,6 +3,8 @@
  * Title: Right Sidebar
  * Slug: purr/right-sidebar
  * Categories: sidebar
```
MaggieCabrera, line 5: this is not one of the [default categories](https://developer.wordpress.org/themes/patterns/registering-patterns/#pattern-categories). Unless we had more than one pattern for it, we shouldn't create new categories

shail-mehta: We have 2 patterns for sidebar left-sidebar & right-sidebar

Left Sidebar has already sidebar category added, so I need to remove sidebar categories in left-sidebar.php pattern also?

MaggieCabrera: mmh, maybe in this theme it makes sense to have a sidebar category, but to have it show up in the pattern inserter we need to [register](https://developer.wordpress.org/themes/patterns/registering-patterns/#registering-a-pattern-category) it first if it's not one of the default ones

shail-mehta: Right, I will Register sidebar category first.

shail-mehta: I have created functions.php to register sidebar Pattern Category.

shail-mehta: sidebar-pattern-category

purr/patterns/right-sidebar.php
```diff
@@ -0,0 +1,17 @@
+<?php
+/**
+ * Title: Right Sidebar
+ * Slug: purr/right-sidebar
+ * Categories: sidebar
+ * Block Types: core/template-part/sidebar
+ * Description: Right Sidebar
```
MaggieCabrera, line 7:
```suggestion
 * Description: Right Sidebar
 * Viewport width: 300
```

MaggieCabrera: This makes the preview look much better:

before | after
--- | ---
Screenshot 2024-06-25 at 12 18 53 | Screenshot 2024-06-25 at 12 19 45

MaggieCabrera (changes requested):

purr/patterns/right-sidebar.php
shail-mehta: Added View port in Right Sidebar Pattern.

MaggieCabrera (approved):
This is looking good, thanks for all your work on this!

PR #220, opened 2024-06-24
Update to Node 20

ajlende:
Node 20 is used by both Gutenberg and WordPress now, so we should use the same version here.


MaggieCabrera (approved):
Thank you for this, I've been wanting to do this. I went ahead and added a .nvmrc file for convenience

PR #222, opened 2024-06-26
Merge of the 146 and 142 themes

iamsam2e:
Combination of the two branches with some slight adjustment based off of feedback and my own assessment confirmed working on local

### DIFF
--- a/poetry/patterns/single.php
+++ b/poetry/patterns/single.php
@@ -1,0 +1,58 @@
+<?php
+
+/**
+ * Title: single
+ * Slug: poetry/single
+ * Categories: hidden
+ * Inserter: no
+ */
+?>
+<!-- wp:template-part {"slug":"header","tagName":"header"} /-->
+
+<!-- wp:group {"layout":{"type":"constrained"}} -->
+<div class="wp-block-group"><!-- wp:group {"style":{"spacing":{"padding":{"top":"96px","right":"0","bottom":"96px","left":"0"},"blockGap":"40px"}},"layout":{"type":"flex","flexWrap":"nowrap","orientation":"horizontal","verticalAlignment":"top","justifyContent":"center"}} -->
+    <div class="wp-block-group" style="padding-top:96px;padding-right:0;padding-bottom:96px;padding-left:0"><!-- wp:group {"style":{"spacing":{"padding":{"top":"96px"}}},"layout":{"type":"default"}} -->
+        <div class="wp-block-group" style="padding-top:96px"><!-- wp:post-date /--></div>
+        <!-- /wp:group -->
+
+        <!-- wp:group {"style":{"spacing":{"padding":{"right":"56px","left":"56px"},"blockGap":"32px"},"border":{"left":{"width":"2px"}}},"layout":{"type":"flex","orientation":"vertical","flexWrap":"nowrap"}} -->
+        <div class="wp-block-group" style="border-left-width:2px;padding-right:56px;padding-left:56px"><!-- wp:image {"id":15,"width":"56px","height":"56px","sizeSlug":"full","linkDestination":"none"} -->
+            <figure class="wp-block-image size-full is-resized"><img src="<?php echo esc_url(get_stylesheet_directory_uri()); ?>/assets/images/feather-small.png" alt="" class="wp-image-15" style="width:56px;height:56px" /></figure>
+
+            <!-- /wp:image -->
+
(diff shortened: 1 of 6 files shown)


poetry/patterns/single.php
```diff
@@ -0,0 +1,58 @@
+<?php
+
+/**
+ * Title: single
+ * Slug: poetry/single
+ * Categories: hidden
```
MaggieCabrera: we don't add categories if the pattern is hidden (it doesn't exist as a default category and if we wanted to create one, we would have to register it first)

poetry/patterns/single.php
```diff
@@ -0,0 +1,58 @@
...
+        <div class="wp-block-group" style="border-left-width:2px;padding-right:56px;padding-left:56px"><!-- wp:image {"id":15,"width":"56px","height":"56px","sizeSlug":"full","linkDestination":"none"} -->
+            <figure class="wp-block-image size-full is-resized"><img src="<?php echo esc_url(get_stylesheet_directory_uri()); ?>/assets/images/feather-small.png" alt="" class="wp-image-15" style="width:56px;height:56px" /></figure>
+
+            <!-- /wp:image -->
+
+            <!-- wp:post-title {"style":{"spacing":{"padding":{"right":"24%"}}}} /-->
+
+            <!-- wp:group {"style":{"spacing":{"blockGap":"8px"},"elements":{"link":{"color":{"text":"var:preset|color|vivid-red"}}}},"layout":{"type":"flex","flexWrap":"nowrap"}} -->
+            <div class="wp-block-group has-link-color"><!-- wp:paragraph {"textColor":"custom-contrast"} -->
+                <p class="has-custom-contrast-color has-text-color">By</p>
```
MaggieCabrera, line 27: can you check if running `npm run escape:patterns` on this PR escapes the strings on this pattern? It's not a blocker

poetry/patterns/single.php
```diff
@@ -0,0 +1,58 @@
...
+    <div class="wp-block-group" style="padding-top:96px;padding-right:0;padding-bottom:96px;padding-left:0"><!-- wp:group {"style":{"spacing":{"padding":{"top":"96px"}}},"layout":{"type":"default"}} -->
+        <div class="wp-block-group" style="padding-top:96px"><!-- wp:post-date /--></div>
+        <!-- /wp:group -->
+
+        <!-- wp:group {"style":{"spacing":{"padding":{"right":"56px","left":"56px"},"blockGap":"32px"},"border":{"left":{"width":"2px"}}},"layout":{"type":"flex","orientation":"vertical","flexWrap":"nowrap"}} -->
+        <div class="wp-block-group" style="border-left-width:2px;padding-right:56px;padding-left:56px"><!-- wp:image {"id":15,"width":"56px","height":"56px","sizeSlug":"full","linkDestination":"none"} -->
+            <figure class="wp-block-image size-full is-resized"><img src="<?php echo esc_url(get_stylesheet_directory_uri()); ?>/assets/images/feather-small.png" alt="" class="wp-image-15" style="width:56px;height:56px" /></figure>
```
MaggieCabrera: Check the [guidelines for pattern creation](https://github.com/WordPress/community-themes?tab=readme-ov-file#pattern-creation-guidelines):

> To create dynamic image links in your block patterns, utilize the `get_template_directory_uri()` function. This function retrieves the URL of the current theme's directory, ensuring that the image links are relative to the theme and work correctly even if the website's directory structure changes or if we are using a child theme. This is essential for maintaining the stability and portability of your patterns.

poetry/patterns/single.php
```diff
@@ -0,0 +1,58 @@
...
+                <!-- wp:post-author {"showAvatar":false,"showBio":false,"textColor":"custom-contrast"} /-->
+            </div>
+            <!-- /wp:group -->
+
+            <!-- wp:post-content /-->
+        </div>
+        <!-- /wp:group -->
+    </div>
+    <!-- /wp:group -->
+
```
MaggieCabrera, line 39: I would expect to see the comments block somewhere around here for the single posts template

MaggieCabrera:
I like the idea that the page would show a featured image and no longer has the left column, but I don't think singular is the right template for it. If we look at the [template hierarchy](https://developer.wordpress.org/themes/basics/template-hierarchy/) I think page.html makes more sense and it's something users are more used to

MaggieCabrera:
I would like to see https://github.com/WordPress/community-themes/pull/179 merged and have that header used for this templates, since the header we are using is way too big and probably should only be used for the home page

iamsam2e:
Actioning this today and tomorrow :D

PR #225, opened 2024-06-30
Fixed Header Alignment & Spacing in Term Theme

### LINKED ISSUE #214: Term: Fix header to fit design
The header looks weird when we have a site logo:

Screenshot 2024-06-18 at 17 12 26

Let's fix the alignment and tweak the header so it looks more like [figma](https://www.figma.com/community/file/1263420783737936752)

shail-mehta:
Fixes: https://github.com/WordPress/community-themes/issues/214


MaggieCabrera:
This could be a solution, but the padding is not just missing for the header, but the overall site. Instead of adding padding everywhere, it's a better solution to add global padding in the Global Styles panel, under Layout

shail-mehta:
> This could be a solution, but the padding is not just missing for the header, but the overall site. Instead of adding padding everywhere, it's a better solution to add global padding in the Global Styles panel, under Layout

I have added global padding in theme.json

PR #226, opened 2024-07-08
Archivist: Updates readme and ignore files

karmatosed:
This updates the name in readme and adds a .gitignore file.

### DIFF
--- a/archivist/.gitignore
+++ b/archivist/.gitignore
@@ -1,0 +1,2 @@
+# General
+.DS_Store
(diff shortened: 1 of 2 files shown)


archivist/.gitignore
```diff
@@ -0,0 +1,2 @@
+# General
+.DS_Store
```
MaggieCabrera, line 2: We already have this for the whole repo, I'm not sure about shipping one specific for the theme

karmatosed: Ah ok happy to remove it then.

PR #232, opened 2024-08-16
Theme preview workflow: use external github action

vcanales:
This proposes using an external copy of the preview themes workflow, which will allow us to run it on `pull_request_target` safely, since the code for the workflow itself lives outside of the repo — this means that the workflow will be able to comment on PRs created by non-members of the WordPress GitHub org.

The code is currently under my GitHub account, but I'll be moving to change ownership to the WordPress org eventually.

### DIFF
--- a/.github/workflows/preview-theme.yml
+++ b/.github/workflows/preview-theme.yml
@@ -16,7 +16,7 @@ jobs:
           ref: ${{ github.event.pull_request.head.sha }}
 
       - name: Preview Theme Changes
-        uses: vcanales/action-wp-playground-pr-preview@trunk
+        uses: WordPress/action-wp-playground-pr-preview@trunk
         with: 
           github-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


MaggieCabrera:
@mikachan did we implement this on the a8c themes? I can't remember if it had security issues

mikachan:
Yes, I think using the external copy worked around the security issues and we were able to use `pull_request_target`: https://github.com/Automattic/themes/pull/8011.

MaggieCabrera:
Awesome, I'll review this asap then. And it would be great to bring the theme check action too to this repo

vcanales:
Keep in mind that the action has been moved to https://github.com/WordPress/action-wp-playground-pr-preview

mikachan:
Thanks @vcanales! I went ahead and updated the action ref: https://github.com/WordPress/community-themes/pull/232/commits/892208b8a63c9929840b3a2fef074339ebad5a75.

PR #235, opened 2024-09-13
Poetry: add header small pattern

### LINKED ISSUE #174: Poetry: Create alternative header

adrianaleites:
- Add spacing sizes to the theme.json as in Twenty Twenty-Four theme.
- Replace spacing with pixels values with spacing presets in header pattern.
- Convert "Food for thoughts" to Heading 1, instead of Paragraph.
- Convert "A collection of valuable thoughts by a very egocentric person." to Site Tagline block, instead of Paragraph.
- Add header small pattern (note: added custom style in theme.json to define the align-items as baseline).
- Replace header in 404, search and archive templates to use instead header small.

Closes #174 
Replaces #179

### DIFF
--- a/poetry/parts/header-small.html
+++ b/poetry/parts/header-small.html
@@ -1,0 +1,1 @@
+<!-- wp:pattern {"slug":"poetry/header-small"} /-->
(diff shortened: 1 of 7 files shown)


MaggieCabrera:
This looks good, thank you! I removed the custom class, since the visible change is very small

PR #253, opened 2025-02-04
Update screenshot.png

bijayyadav:


kafleg:
@bijayyadav Would you please change the site title from **Block FSE** to **Darken**

kafleg (approved):

PR #263, opened 2025-12-17
Add a test theme for the Nav Overlay

MaggieCabrera:
Do NOT merge

This is a test theme to test the new overlay feature. Best tested with https://github.com/WordPress/gutenberg/pull/74069 checked out. The theme is just a blank empty theme with the following changes:

- Added the overlay.html template part. This doesn't work yet, this PR will be helpful to test that feature. This part uses the overlay 1 pattern
- Added two patterns for the overlay. Both should show on the sidebar when creating a new overlay pattern

Screenshot 2025-12-17 at 16 02 50


overlay-test/parts/navigation-overlay.html
```diff
@@ -0,0 +1 @@
+<!-- wp:pattern {"slug":"overlay-test/overlay-1"} /-->
```
MaggieCabrera, line 1: This is not showing up yet as the default overlay, it's expecting a user created one

overlay-test/patterns/overlay-1.php
```diff
@@ -0,0 +1,18 @@
+<?php
+/**
+ * Title: Navigation Overlay 1
+ * Slug: overlay-test/overlay-1
+ * Categories: navigation
+ * Block Types: core/template-part/overlay
```
MaggieCabrera, line 6: This line is what makes it show on the right sidebar

overlay-test/patterns/overlay-1.php
```diff
@@ -0,0 +1,18 @@
...
+?>
+
+<!-- wp:group {"style":{"spacing":{"padding":{"right":"var:preset|spacing|40","left":"var:preset|spacing|40","top":"var:preset|spacing|40","bottom":"var:preset|spacing|40"}}},"backgroundColor":"cyan-bluish-gray","layout":{"type":"flex","orientation":"vertical","flexWrap":"wrap"}} -->
+<div class="wp-block-group has-cyan-bluish-gray-background-color has-background" style="padding-top:var(--wp--preset--spacing--40);padding-right:var(--wp--preset--spacing--40);padding-bottom:var(--wp--preset--spacing--40);padding-left:var(--wp--preset--spacing--40)"><!-- wp:navigation-overlay-close -->
+<button class="wp-block-navigation-overlay-close" type="button" aria-label="Close"><svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="24" height="24" aria-hidden="true" focusable="false"><path d="M13 11.8l6.1-6.3-1.1-1-6.1 6.2-6.1-6.2-1.1 1 6.1 6.3-6.5 6.7 1.1 1 6.5-6.6 6.5 6.6 1.1-1z"></path></svg></button>
+<!-- /wp:navigation-overlay-close -->
+<!-- wp:heading -->
+<h2 class="wp-block-heading">This is the default Overlay for this theme</h2>
+<!-- /wp:heading -->
+<!-- wp:navigation {"layout":{"type":"flex","orientation":"vertical"}} /--></div>
```
MaggieCabrera, line 17: The Nav block will show the same as what the header nav block shows

overlay-test/parts/navigation-overlay.html
jeryj: Even if it did, would we a way to get the parent navigation block it's assigned to? Maybe in the render we'd know to grab the parent navigation if this is the templating style?

overlay-test/patterns/overlay-1.php
MaggieCabrera: @getdave this will be the pattern we show on the theme's tempalte part

overlay-test/parts/navigation-overlay.html
MaggieCabrera: @getdave it's just a one liner but this is ok, we do this often. The template part invokes the pattern

overlay-test/parts/header.html
```diff
@@ -0,0 +1,30 @@
...
+			<!-- wp:group -->
+			<div class="wp-block-group">
+				<!-- wp:site-title /-->
+				<!-- wp:site-tagline /-->
+			</div>
+			<!-- /wp:group -->
+		</div>
+		<!-- /wp:group -->
+	
+		<!-- wp:navigation {"overlay":"navigation-overlay"} /-->
```
MaggieCabrera, line 19: @getdave this is the key change, we are telling the nav block to use a particular overlay that we have defined before

overlay-test/theme.json
```diff
@@ -0,0 +1,45 @@
...
+		{
+			"area": "header",
+			"name": "header"
+		},
+		{
+			"area": "footer",
+			"name": "footer"
+		},
+		{
+			"area": "navigation-overlay",
```
MaggieCabrera, line 41: @getdave we also need to register the template part here, like we do header and footer

overlay-test/parts/navigation-overlay.html
getdave: Yes it's a good approach and keeps things DRY.

PR #265, opened 2026-03-06
Add wp-env setup for this repo

MaggieCabrera:
Adds a local development environment powered by wp-env. All themes in the repo are available out of the box.

 - Auto-detects a free port starting from 8888, avoiding conflicts with other services
  - `--theme <slug>` activates a theme after startup; if the environment is already running, it skips startup
  and just switches the active theme
  - `env:stop` and `env:destroy` read the saved port from .wp-env-port to always target the correct instance
  - Theme activation is verified against WP-CLI output. A success message is only shown when the theme was
  actually switched

  How to test
  1. Fresh start
  npm run env:start -- --theme blue-note
  - Visit http://localhost:8888, log in (admin / password)
  - Confirm Blue Note is the active theme

  2. Switch theme on a running env
  npm run env:start -- --theme archivist
  - Should skip startup, just switch the theme
  - Confirm Archivist is now active (no restart required)

  3. Invalid theme slug
  npm run env:start -- --theme this-does-not-exist
  - Should print an error and exit with a non-zero code
  - Should NOT show a success message

  4. Start with no theme flag
  npm run env:stop
  npm run env:start
  - Should start normally, no theme activation step

  5. Stop and restart with a different theme
  npm run env:stop
  npm run env:start -- --theme blue-note
  - Should start the existing environment (not create a new one)
  - Confirm Blue Note is active at http://localhost:8888

  6. Destroy
  npm run env:destroy
  - Containers and volumes removed, .wp-env-port file deleted
  - Running npm run env:start after this should boot a clean instance

### DIFF
--- a/.wp-env.gutenberg.json
+++ b/.wp-env.gutenberg.json
@@ -1,0 +1,8 @@
+{
+	"core": "WordPress/WordPress",
+	"plugins": [ "https://downloads.wordpress.org/plugin/gutenberg.latest-stable.zip" ],
+	"themes": [ "https://downloads.wordpress.org/theme/twentytwentythree.latest-stable.zip" ],
+	"mappings": {
+		"wp-content/themes": "."
+	}
+}
--- a/env.mjs
+++ b/env.mjs
@@ -18,20 +18,16 @@ import { readFileSync, writeFileSync, existsSync, unlinkSync } from 'fs';
 
 const PORT_FILE = '.wp-env-port';
 
-function findFreePort( start = 8888 ) {
+function isPortFree( port ) {
 	return new Promise( ( resolve ) => {
 		const server = createServer();
-		server.on( 'error', () => resolve( findFreePort( start + 1 ) ) );
-		server.listen( start, () => server.close( () => resolve( start ) ) );
+		server.on( 'error', () => resolve( false ) );
+		server.listen( port, () => server.close( () => resolve( true ) ) );
 	} );
 }
 
-function isPortInUse( port ) {
-	return new Promise( ( resolve ) => {
-		const server = createServer();
-		server.on( 'error', () => resolve( true ) );
-		server.listen( port, () => server.close( () => resolve( false ) ) );
-	} );
+async function findFreePort( start = 8888 ) {
+	return ( await isPortFree( start ) ) ? start : findFreePort( start + 1 );
 }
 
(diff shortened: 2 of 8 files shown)


.wp-env.gutenberg.json
```diff
@@ -0,0 +1,8 @@
+{
+	"core": "WordPress/WordPress",
+	"plugins": [ "https://downloads.wordpress.org/plugin/gutenberg.latest-stable.zip" ],
+	"themes": [ "https://downloads.wordpress.org/theme/twentytwentythree.latest-stable.zip" ],
```
MaggieCabrera, line 4: Atlas needs TT3 as a dependency

env.mjs
```diff
@@ -0,0 +1,52 @@
...
+const args = process.argv.slice( 2 );
+const themeIndex = args.indexOf( '--theme' );
+let themeSlug = null;
+if ( themeIndex !== -1 ) {
+	themeSlug = args[ themeIndex + 1 ];
+	args.splice( themeIndex, 2 );
+}
+
+const port = await findFreePort();
+const testsPort = await findFreePort( port + 1 );
```
mikachan: These port checks are called for `env:start`, but we don't use anything similar for `env:stop` or `env:destroy`. Should the port that's being used be checked for these commands? I'm wondering if `wp-env` might accidentally stop or destroy the wrong instance if it hasn't checked the port.

mikachan (commented):
When I ran `npm run env:start -- --theme blue-note`, the env started fine, and I saw this success:

Image

But it tried to activate Twenty Twenty-Five instead, which doesn't exist here. I could still manually switch to Blue Note afterwards, which worked fine.

And when I stopped the env and then ran `npm run env:start -- --theme archivist`, I saw the same success message again:

Image

But Blue Note was still active.

Otherwise, the test environment is working well for me.

env.mjs
MaggieCabrera: I think this whole thing might be more trouble than anything, I'll see what I can do

MaggieCabrera: I'm making it so that we store the active port in a file, and if we re run the env with a new theme it just changes it instead of destroying it/starting a new one again.

mikachan (approved):
Running `npm run env:start -- --theme blue-note` is working well for me now. Thanks! 🚢

